PR #36, opened 2026-07-18
fix: free Info via PMIx_Info_free; stop forgetting CStrings

### LINKED ISSUE #21: Info never frees PMIx_Info_create allocations (memory leak)
## Priority
HIGH

## Problem
`Info` holds a `*mut pmix_info_t` from `PMIx_Info_create` (`InfoBuilder::build`, `info_with_string_key`) but has **no `Drop` impl**. Every constructed `Info` leaks the C allocation for the process lifetime.

## Location
- `src/lib.rs` — `pub struct Info`, `InfoBuilder::build`, `info_with_string_key`

## Fix
Implement `Drop for Info` using the correct PMIx free path for the linked version (`PMIx_Info_free` / `PMIX_INFO_FREE` / destruct+free loop). Add regression tests.

## Source
REVIEW.md (2026-07-18) — N1

### LINKED ISSUE #22: info_with_string_key mem::forget of CString key/value is unsafe/unportable
## Priority
HIGH

## Problem
`info_with_string_key` calls `std::mem::forget` on key/value `CString`s after `PMIx_Info_load`, assuming PMIx always copies. If the library stores pointers, this is UAF; if it copies, Rust still permanently leaks the CStrings.

## Location
- `src/lib.rs` — `info_with_string_key`

## Fix
Keep CStrings alive only for the load call if PMIx copies; verify with ASan against the supported PMIx versions. Document the contract.

## Source
REVIEW.md (2026-07-18) — N2

SedahsDev:
## Summary
Addresses review comments on #21 and #22 (fix together).

- `Drop for Info` calls `PMIx_Info_free(handle, len)`
- `InfoBuilder::build` frees on load failure / null create
- `info_with_string_key`: CStrings live only across `PMIx_Info_load` (PMIx copies), then drop — **no** `mem::forget`
- `Info::into_raw` escape hatch for callers that take ownership
- Unit tests for drop / into_raw

**Note on other `mem::forget` sites** (payload byte objects / data arrays): those intentionally transfer ownership into `pmix_value_t` freed later by `free_value` / `PmixOwnedValue::Drop` — not the same bug as the Info CString forget.

Fixes #21
Fixes #22

## Test plan
- [x] `cargo test --lib test_info_`


SedahsDev (commented):
LGTM — solid fix. The `Drop for Info` implementation is correct:

- Null check on handle before freeing prevents double-free
- `into_raw` escape hatch with `mem::forget` is the right pattern for ownership transfer
- `info_with_string_key` no longer leaks CStrings — they drop naturally after `PMIx_Info_load` copies
- Null check on `PMIx_Info_create` return value is good practice
- Free on load failure in `InfoBuilder::build` prevents leaking partial arrays

One minor note: the `panic!` on `PMIx_Info_create` returning null could be converted to a `Result` in the future for callers who want graceful error handling, but panic is reasonable for now given that `PMIx_Info_create` should almost never fail.

PR #37, opened 2026-07-18
fix: log finalize errors in Context::Drop (no panic)

### LINKED ISSUE #23: Context::Drop panics via finalize(None).unwrap()
## Priority
HIGH

## Problem
```rust
impl Drop for Context {
    fn drop(&mut self) {
        finalize(None).unwrap();
    }
}
```
Panicking in `Drop` is dangerous (double-panic abort, hostile to multi-init edge cases). Finalize can legitimately fail.

## Location
- `src/lib.rs` — `impl Drop for Context`

## Fix
Ignore or log finalize errors in Drop; never unwrap.

## Source
REVIEW.md (2026-07-18) — N3

SedahsDev:
## Summary
Per issue comments: do not unwrap in Drop; **log even on release builds**.

```rust
if let Err(status) = finalize(None) {
    eprintln!("pmix: finalize in Context::Drop failed: status={status}");
}
```

Fixes #23

## Test plan


SedahsDev (commented):
Clean fix. Logging finalize errors in `Context::Drop` with `eprintln!` instead of panicking is the right approach. The comment explaining "never panic in Drop" is good documentation for future maintainers.

One consideration: in HPC environments with thousands of processes, `eprintln!` could create a lot of stderr noise during batch job teardown. If that becomes an issue, consider gating the log behind `log::warn!` or `tracing::warn!` with a feature flag, but the current approach is fine for now.

PR #38, opened 2026-07-18
fix: feature-gate mock_ffi and cfg-split mock dispatch

### LINKED ISSUE #24: mock_ffi is public in non-test builds without a Cargo feature
## Priority
HIGH

## Problem
`src/lib.rs` has `pub mod mock_ffi` unconditionally. Test doubles are part of the public API surface. Production modules use `cfg!(any(test, feature = \"mock_ffi\"))` but `Cargo.toml` does not define a `mock_ffi` feature.

## Location
- `src/lib.rs`
- `src/mock_ffi.rs`
- call sites in server/data_ops/etc.

## Fix
- Gate with `#[cfg(any(test, feature = \"mock_ffi\"))]`
- Declare optional feature in Cargo.toml
- Document that mocks are test-only

## Source
REVIEW.md (2026-07-18) — N4

SedahsDev:
## Summary
Per #24 comments: gate the module, declare the feature, audit test helpers.

- Cargo feature `mock_ffi = []`
- `#[cfg(any(test, feature = \"mock_ffi\"))] pub mod mock_ffi`
- server/fabric mock branches rewritten with `#[cfg]` so non-feature builds never typecheck mock symbols
- Document **`MockGuard`** as the preferred RAII helper (covers enable/disable + panic paths); `enable_mock_ffi` / `disable_mock_ffi` remain for low-level control
- Macro `pmix_ffi_or_mock!` available for future call sites

Fixes #24

## Test plan
- [x] `cargo build` (no feature)
- [x] `cargo build --features mock_ffi`


SedahsDev (commented):
Good feature gate implementation. The `#[cfg(any(test, feature = "mock_ffi"))]` pattern is correct and the `pmix_ffi_or_mock!` macro is a clean dispatch mechanism.

Two notes:
1. The `default = []` in `[features]` is fine but could be omitted since empty defaults are the implicit behavior — just a style nit.
2. The `MockGuard` RAII helper mentioned in the PR description is a nice touch for test ergonomics — it covers enable/disable + panic paths cleanly.

This properly closes the gap where production builds were shipping test doubles unconditionally.

PR #41, opened 2026-07-18
feat: finalize PMIx server on PmixServerHandle Drop (RAII)

### LINKED ISSUE #29: PmixServerHandle should RAII-finalize or be #[must_use] with clear guard API
## Priority
MEDIUM

## Problem
Docs state dropping `PmixServerHandle` does not call `server_finalize`. Easy to leak server resources if callers forget finalize.

## Fix
Prefer Drop-finalize with `mem::forget` escape, or `#[must_use]` guard type requiring explicit finalize.

## Source
REVIEW.md (2026-07-18) — N9

SedahsDev:
## Summary
Per #29 Option A (recommended), aligned with #23 Context::Drop:
- Drop calls `PMIx_server_finalize` and logs failures (including release)
- `server_finalize` disarms Drop (no double-finalize)
- `into_raw` / `mem::forget` escape hatch
- `#[must_use]` on the handle type

Fixes #29

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -269,17 +269,53 @@ impl PmixServerModule {
 
 /// RAII handle returned by [`server_init`].
 ///
-/// Dropping this handle does **not** automatically call
-/// `PMIx_server_finalize` — the server must be explicitly finalized
-/// via [`server_finalize`] to ensure proper cleanup of internal
-/// communication channels and memory.
+/// Dropping this handle **finalizes** the PMIx server library (same path as
+/// [`server_finalize`]). This matches idiomatic Rust handle types and keeps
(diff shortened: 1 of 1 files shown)


SedahsDev (commented):
Solid RAII implementation. The `active` flag pattern is clean — `server_finalize` disarms Drop by setting `active = false` before `mem::forget`, preventing double-finalize. `into_raw` provides the escape hatch for callers who need manual control.

The `#[must_use]` on `PmixServerHandle` is a good safety net — callers who drop the handle without using it get a warning.

One thing to verify: `ffi::PMIx_server_finalize()` is called directly here rather than going through `server_finalize()`. This is intentional since `server_finalize` would disarm Drop and we're already in Drop. Just noting it for clarity.

PR #364, opened 2026-08-03
threading: callback hop-off-progress helpers + bridge policy (#51)

### LINKED ISSUE #51: threading: callback hop-off-progress helpers + bridge policy
## Phase 4a — Callbacks (refined)

OpenPMIx delivers `_nb` completions and events on the **progress thread**, not the caller.

### Rules
- Must **not** call blocking PMIx APIs from the handler without hopping to an app thread
- Non-blocking from handler only if you do not wait in-handler
- Bridges stay minimal; **do not** hold registry `Mutex` across user callback execution
- cbdata uses `crate::cbdata::encode_req_id` / `decode_req_id` (done in data_ops/query_log/security/server)

### Deliver
1. Shared helper: channel / `spawn_from_callback` template
2. Docs + example (get_nb or fence_nb + one events path)
3. Optional ZST `ProgressContext` marker documenting forbidden APIs
4. Audit remaining bridges for lock-hold-across-user-code

### Acceptance
- [ ] Helper used by ≥1 data_ops nb + ≥1 events path
- [ ] Docs in module + THREADING.md cross-link
- [ ] Deadlock regression (timeout) if feasible

### Depends on
Client session (#48) helpful for examples.

SedahsDev:
## Summary

Implements #51 (Phase 4a — callbacks). OpenPMIx delivers `_nb` completions and events on the **progress thread**, never the caller. This PR ships the shared hop-off helpers, a documented bridge policy, and brings the last out-of-policy bridge (`events`) in line.

### 1. Shared helper — `src/threading.rs` (new `pub mod threading`)
- `ProgressContext` — ZST marker documenting the APIs forbidden on the progress thread (blocking data ops, blocking events, session lifecycle, `progress()`).
- `spawn_from_callback(f)` — fire-and-forget application thread for blocking work from inside a callback.
- `CallbackChannel<T>` — app-thread receiver + clone-able sender for callbacks; documents the `!Send` C-owned value → Rust-owned (`bytes_copy()`) conversion.
- Module docs include the full hop-off template and the 4-point bridge policy (no blocking PMIx in-handler, no waiting in-handler, no registry `Mutex` across user code, `cbdata::encode/decode_req_id`).

### 2. `events` bridge rework (the remaining out-of-policy bridge)
The old bridge boxed the user `NotificationFn` into `cbdata` and **freed it immediately after registration** — a use-after-free, since PMIx retains the handler (and its cbdata) for the lifetime of the registration. Replaced with:
- `HANDLER_REGISTRY`: ref-ID-keyed `Box<NotificationFn>`, freed at deregistration.
- `notification_bridge` resolves the user fn under a scoped lock, released **before** the user callback runs.
- `register_event_handler_nb` parks the fn when the ref ID arrives via the completion bridge (previously leaked); error path frees it.
- `deregister_event_handler` removes the entry before the C call (no fire-into-freed-memory window).

### 3. Docs + example
- `examples/callback_hop.rs`: uses the helpers with **one data_ops nb path** (`get_nb`) **and one events path** (`register_event_handler` + `notify_event`) — the issue's acceptance criterion. Runs standalone with a graceful no-DVM exit.
- `THREADING.md` + crate-root concurrency docs cross-link the module; roadmap marks #51 Done.

### 4. Audit (deliverable 4)
Checked every `extern "C"` bridge: `data_ops`, `query_log`, `security`, `monitoring` already follow the policy (lock scoped to registry removal, user code outside the lock). `events` was the exception and is now fixed, with a **deadlock regression test** (`test_notification_bridge_no_lock_across_user_code`) proving the lock is not held across user callback execution.

## Test plan
- `cargo build --examples` ✅
- `cargo test --lib`: **1754 passed, 0 failed** (incl. 8 new `threading` + 121 `events` tests, 3 doctests)
- `cargo clippy --lib`: no new warnings
- `examples/callback_hop` runs standalone (graceful no-DVM exit)

Note: the integration-test suite under `tests/` has a pre-existing breakage on `main` (`tests/daemon_helper.rs` still uses the pre-#70 `PmixToolHandle` API) — unrelated to this PR. CI for this repo has no recorded workflow runs.

Closes #51

### DIFF
--- a/examples/callback_hop.rs
+++ b/examples/callback_hop.rs
@@ -127,7 +127,7 @@ fn main() {
 /// OpenPMIx requires every handler to complete the event chain by calling
 /// `cbfunc(..., cbdata)`. `progress_local_event_hdlr` thread-shifts, so the
 /// completion may safely run from the hopped-off application thread.
-unsafe extern "C" fn on_event(
+unsafe fn on_event(
     id: pmix::events::EventHandlerRef,
     status: i32,
     _source: *const c_void,
--- a/examples/callback_hop.rs
+++ b/examples/callback_hop.rs
@@ -1,0 +1,159 @@
+//! Callback hop-off-progress example (issue #51).
+//!
+//! Demonstrates the two shared helpers from [`pmix::threading`] with **one
+//! `data_ops` non-blocking path** (`get_nb`) and **one events path**
+//! (`register_event_handler` + `notify_event`):
+//!
+//! * [`CallbackChannel`](pmix::threading::CallbackChannel) — the application
+//!   thread keeps the receiver; the `get_nb` completion callback (which PMIx
+//!   invokes on the **progress thread**) pushes Rust-owned data through a
+//!   cloned sender and returns immediately.
+//! * [`spawn_from_callback`](pmix::threading::spawn_from_callback) — the event
+//!   handler (also on the progress thread) hops blocking work onto a fresh
+//!   application thread.
+//! * [`ProgressContext`](pmix::threading::ProgressContext) — a zero-sized
+//!   marker that documents "we are on the progress thread" and the forbidden
+//!   APIs at the point of use.
+//!
+//! # Running
+//!
+//! The example compiles and runs standalone (it exits gracefully when no DVM
+//! is present). Under a process manager it exercises real callbacks:
+//!
+//! ```text
+//! prterun --np 2 ./target/debug/examples/callback_hop
+//! ```
+
+use std::os::raw::c_void;
+use std::sync::mpsc;
+
+fn main() {
+    println!("pmix-rs: callback hop-off-progress example (issue #51)");
+    println!("Run under a DVM to see real callbacks, e.g.");
+    println!("  prterun --np 2 ./target/debug/examples/callback_hop");
+
+    // Without a DVM there is no PMIx server to connect to — exit gracefully
+    // like the other role examples in this crate.
+    let client = match pmix::PmixClient::connect_new(None) {
+        Ok(c) => c,
+        Err(e) => {
+            eprintln!("connect_new failed (need prterun/DVM?): {e:?}");
+            return;
+        }
+    };
+    let proc = client.require_proc();
+    println!("connected: rank {}", proc.get_rank());
+
+    // ── 1. Events path ─────────────────────────────────────────────────────
+    // Empty `codes` = handle all events. The handler runs on the PMIx
+    // progress thread, so it hops off before doing any work.
+    let info = pmix::info::empty();
+    let handler_ref = match pmix::events::register_event_handler(&[], &info, Some(on_event), None) {
+        Ok(r) => r,
+        Err(e) => {
+            println!("register_event_handler failed (no DVM?): {e:?}");
+            0
+        }
+    };
+
+    // ── 2. data_ops non-blocking path ──────────────────────────────────────
+    // Application thread owns the channel; the callback gets a cloned sender.
+    let hop = pmix::threading::CallbackChannel::<(i32, Option<Vec<u8>>)>::new();
+    let tx = hop.sender();
+    match pmix::data_ops::get_nb(&proc, "pmix.job.size", None, Box::new(HopCallback { tx })) {
+        Ok(()) => println!("get_nb accepted — completion will arrive on the progress thread"),
+        Err(e) => println!("get_nb submit failed (no DVM?): {e:?}"),
+    }
+
+    // ── 3. Deliver an event so the handler fires (needs a DVM) ─────────────
+    let event = pmix::PmixStatus::Known(pmix::PmixError::EventJobEnd);
+    match pmix::events::notify_event(event, &proc, pmix::PmixDataRange::Session, &info) {
+        Ok(()) => println!("notify_event accepted — handler will fire on the progress thread"),
+        Err(e) => println!("notify_event failed (no DVM?): {e:?}"),
+    }
+
+    // ── 4. Application thread: drain the hop channel ───────────────────────
+    // This thread is where blocking PMIx calls are legal (the callback that
+    // pushed into the channel must never block on progress).
+    let deadline = std::time::Instant::now() + std::time::Duration::from_secs(5);
+    loop {
+        match hop.recv_timeout(std::time::Duration::from_millis(250)) {
+            Ok((status, payload)) => {
+                let text = payload
+                    .as_deref()
+                    .map(|b| String::from_utf8_lossy(b).into_owned())
+                    .unwrap_or_else(|| "<no value>".to_string());
+                println!("[app thread] get_nb completed: status={status} value={text:?}");
+                break;
+            }
+            Err(mpsc::RecvTimeoutError::Timeout) if std::time::Instant::now() < deadline => {
+                // Standalone runs never produce the completion; keep waiting
+                // briefly so a real DVM run has time to deliver it.
+            }
+            Err(_) => {
+                println!("[app thread] no get_nb completion within 5s (standalone run?)");
+                break;
+            }
+        }
+    }
+
+    // ── 5. Cleanup ─────────────────────────────────────────────────────────
+    if handler_ref != 0 {
+        match pmix::events::deregister_event_handler(handler_ref, None) {
+            Ok(()) => println!("deregistered event handler {handler_ref}"),
+            Err(e) => println!("deregister_event_handler: {e:?}"),
+        }
+    }
+    let _ = client.disconnect(None);
+    println!("callback_hop example done");
+}
+
+/// Event handler — runs on the PMIx **progress thread**.
+///
+/// Note the [`ProgressContext`](pmix::threading::ProgressContext) marker: it
+/// documents that blocking PMIx calls are forbidden here, and the work is
+/// hopped onto a fresh application thread via
+/// [`spawn_from_callback`](pmix::threading::spawn_from_callback).
+unsafe extern "C" fn on_event(
+    id: pmix::events::EventHandlerRef,
+    status: i32,
+    _source: *const c_void,
+    _info: *mut c_void,
+    _ninfo: usize,
+    _results: *mut c_void,
+    _nresults: usize,
+    _cbfunc: pmix::events::pmix_event_notification_cbfunc_fn_t,
+    _cbdata: *mut c_void,
+) {
+    // We are on the progress thread — this marker is documentation, and any
+    // blocking PMIx call here would deadlock progress.
+    let _ctx = pmix::threading::ProgressContext;
+
+    // Hop off: spawn an application thread for the actual handling. We never
+    // join it here (that could block progress and deadlock the process).
+    let _ = pmix::threading::spawn_from_callback(move || {
+        println!("[app thread] event {id} fired with status {status} — blocking work is safe here");
+    });
+}
+
+/// `get_nb` completion callback — also invoked on the PMIx progress thread.
+///
+/// The [`PmixOwnedValue`](pmix::PmixOwnedValue) handed to us is a C-owned
+/// handle (`!Send`), so it is converted to Rust-owned bytes with
+/// [`bytes_copy`](pmix::PmixOwnedValue::bytes_copy) **before** crossing the
+/// channel. Only the cheap, non-blocking `send` happens on the progress
+/// thread; the application thread does any blocking work.
+struct HopCallback {
+    tx: mpsc::Sender<(i32, Option<Vec<u8>>)>,
+}
+
+impl pmix::data_ops::GetValueCallback for HopCallback {
+    fn on_result(self: Box<Self>, status: pmix::PmixStatus, value: Option<pmix::PmixOwnedValue>) {
+        let _ctx = pmix::threading::ProgressContext;
+
+        // Convert the C-owned (!Send) value into Rust-owned bytes so it can
+        // cross the channel, then hop — never block here.
+        let payload = value.map(|v| v.bytes_copy());
+        let _ = self.tx.send((status.to_raw(), payload));
+    }
+}
--- a/examples/callback_hop.rs
+++ b/examples/callback_hop.rs
@@ -10,7 +10,7 @@
 //!   cloned sender and returns immediately.
 //! * [`spawn_from_callback`](pmix::threading::spawn_from_callback) — the event
 //!   handler (also on the progress thread) hops blocking work onto a fresh
-//!   application thread.
+//!   application thread, then completes the OpenPMIx event chain via `cbfunc`.
 //! * [`ProgressContext`](pmix::threading::ProgressContext) — a zero-sized
 //!   marker that documents "we are on the progress thread" and the forbidden
 //!   APIs at the point of use.
@@ -25,6 +25,7 @@
 //! ```
 
 use std::os::raw::c_void;
+use std::ptr;
 use std::sync::mpsc;
 
 fn main() {
@@ -47,12 +48,18 @@ fn main() {
     // ── 1. Events path ─────────────────────────────────────────────────────
     // Empty `codes` = handle all events. The handler runs on the PMIx
     // progress thread, so it hops off before doing any work.
+    //
+    // Track success via `Result` — OpenPMIx assigns handler indices from 0, so
+    // `Ok(0)` is a legitimate first registration, not a failure sentinel.
     let info = pmix::info::empty();
     let handler_ref = match pmix::events::register_event_handler(&[], &info, Some(on_event), None) {
-        Ok(r) => r,
+        Ok(r) => {
+            println!("registered event handler {r}");
+            Some(r)
+        }
         Err(e) => {
             println!("register_event_handler failed (no DVM?): {e:?}");
-            0
+            None
         }
     };
 
@@ -66,6 +73,8 @@ fn main() {
     }
 
     // ── 3. Deliver an event so the handler fires (needs a DVM) ─────────────
+    // Blocking notify_event waits on the event chain; the handler must call
+    // its completion cbfunc (see on_event) or this never returns under a DVM.
     let event = pmix::PmixStatus::Known(pmix::PmixError::EventJobEnd);
     match pmix::events::notify_event(event, &proc, pmix::PmixDataRange::Session, &info) {
         Ok(()) => println!("notify_event accepted — handler will fire on the progress thread"),
@@ -98,7 +107,7 @@ fn main() {
     }
 
     // ── 5. Cleanup ─────────────────────────────────────────────────────────
-    if handler_ref != 0 {
+    if let Some(handler_ref) = handler_ref {
         match pmix::events::deregister_event_handler(handler_ref, None) {
             Ok(()) => println!("deregistered event handler {handler_ref}"),
             Err(e) => println!("deregister_event_handler: {e:?}"),
@@ -114,6 +123,10 @@ fn main() {
 /// documents that blocking PMIx calls are forbidden here, and the work is
 /// hopped onto a fresh application thread via
 /// [`spawn_from_callback`](pmix::threading::spawn_from_callback).
+///
+/// OpenPMIx requires every handler to complete the event chain by calling
+/// `cbfunc(..., cbdata)`. `progress_local_event_hdlr` thread-shifts, so the
+/// completion may safely run from the hopped-off application thread.
 unsafe extern "C" fn on_event(
     id: pmix::events::EventHandlerRef,
     status: i32,
@@ -122,8 +135,8 @@ unsafe extern "C" fn on_event(
     _ninfo: usize,
     _results: *mut c_void,
     _nresults: usize,
-    _cbfunc: pmix::events::pmix_event_notification_cbfunc_fn_t,
-    _cbdata: *mut c_void,
+    cbfunc: pmix::events::pmix_event_notification_cbfunc_fn_t,
+    cbdata: *mut c_void,
 ) {
     // We are on the progress thread — this marker is documentation, and any
     // blocking PMIx call here would deadlock progress.
@@ -131,8 +144,30 @@ unsafe extern "C" fn on_event(
 
     // Hop off: spawn an application thread for the actual handling. We never
     // join it here (that could block progress and deadlock the process).
+    // Complete the OpenPMIx event chain from the app thread after the work.
+    //
+    // Raw `*mut c_void` is `!Send`; carry the chain address as `usize` across
+    // the hop. Function pointers are already `Send`. Calling `cbfunc` from the
+    // app thread is the intended hop-off pattern (`progress_local_event_hdlr`
+    // does PMIX_THREADSHIFT internally).
+    let chain_addr = cbdata as usize;
     let _ = pmix::threading::spawn_from_callback(move || {
         println!("[app thread] event {id} fired with status {status} — blocking work is safe here");
+        if let Some(cbfunc) = cbfunc {
+            // SAFETY: complete the event chain; return OpenPMIx's chain pointer
+            // verbatim (never NULL). Address was received from OpenPMIx and is
+            // valid until the chain completes.
+            unsafe {
+                cbfunc(
+                    status,
+                    ptr::null_mut(),
+                    0,
+                    None,
+                    ptr::null_mut(),
+                    chain_addr as *mut c_void,
+                );
+            }
+        }
     });
 }
 
--- a/src/events.rs
+++ b/src/events.rs
@@ -22,6 +22,17 @@
 //! call is blocking and returns the handler reference ID directly in the
 //! return status (positive = success, negative = error).
 //!
+//! # Threading
+//!
+//! Event handlers are delivered on the PMIx **progress thread**, like all
+//! `_nb` completions. The notification bridge follows the policy in
+//! [`crate::threading`]: the user's `NotificationFn` is stored in a registry
+//! keyed by the handler reference ID (never a raw pointer in `cbdata`), and
+//! the registry lock is **not** held while the user callback runs. Handlers
+//! must not call blocking PMIx APIs; hop off with
+//! [`spawn_from_callback`](crate::threading::spawn_from_callback) or a
+//! [`CallbackChannel`](crate::threading::CallbackChannel) first.
+//!
 //! # Example
 //!
 //! ```no_run
@@ -42,9 +53,11 @@
 //! deregister_event_handler(handler_ref, None).expect("deregister failed");
 //! ```
 
-use crate::{Info, PmixDataRange, PmixError, PmixStatus, Proc, ffi};
+use crate::{ffi, Info, PmixDataRange, PmixError, PmixStatus, Proc};
+use std::collections::HashMap;
 use std::os::raw::c_void;
 use std::ptr;
+use std::sync::{LazyLock, Mutex};
 
 /// PMIx event notification completion callback type (re-exported for external use).
 ///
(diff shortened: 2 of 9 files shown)


SedahsDev (commented):
## Review Summary

Overall, this PR is well-structured and addresses the core issue (#51) effectively. The implementation introduces appropriate abstractions for callback hop-off, fixes a critical use-after-free in the events bridge, and includes comprehensive tests and documentation. The approach aligns with the existing patterns in the codebase (data_ops, query_log, etc.) and maintains thread safety.

### Specific Suggestions

#### 1. `threading` module - `spawn_from_callback` panic handling
The fire-and-forget `spawn_from_callback` helper should include panic safety to prevent application crashes from propagating silently. Consider wrapping the closure in `std::panic::catch_unwind`:

```rust
pub fn spawn_from_callback<F>(f: F)
where
    F: FnOnce() + Send + 'static,
{
    std::thread::spawn(move || {
        if let Err(payload) = std::panic::catch_unwind(|| f()) {
            // Log panic or handle appropriately - e.g., using env_logger or tracing
            eprintln!("Callback thread panicked: {:?}", payload);
            // Optionally restart or notify supervisor depending on application requirements
        }
    });
}
```

**Why**: Without panic handling, a panic in the spawned thread will terminate the entire process (since the default panic hook calls `exit()`). This is especially risky in HPC environments where stability is paramount.

#### 2. Events bridge - deregistration safety
While the PR correctly removes the handler from the registry before the C deregister call, consider adding a comment explaining why this ordering is safe (relying on OpenPMIx's guarantee that `Pmix_Deregister_event_handler` prevents future callbacks for that handler). This clarifies the intent for future maintainers.

**Example comment**:
```rust
// SAFE: OpenPMIx guarantees that after Pmix_Deregister_event_handler returns,
// no further callbacks will be delivered for this handler. Removing from
// registry first ensures no application-thread callback attempts to use
// a handler that's being deregistered.
```

#### 3. `CallbackChannel` - sender/receiver symmetry
Verify that the `CallbackChannel` implementation properly handles sender drop (i.e., receiver should not panic when trying to recv on a closed channel). If using `std::sync::mpsc`, the receiver should handle `RecvError` gracefully:

```rust
match receiver.try_recv() {
    Ok(msg) => Some(msg),
    Err(TryRecvError::Empty) => None,
    Err(TryRecvError::Disconnected) => {
        // Channel closed - no more messages will arrive
        None
    }
}
```

**Why**: Prevents unexpected panics if the sender is dropped before the receiver finishes.

#### 4. Documentation - `ProgressContext` examples
In `THREADING.md` or the module docs, consider adding a concrete example showing forbidden operations on the progress thread:

```markdown
### Forbidden Operations on Progress Thread
```rust,ignore
// ❌ NEVER DO THIS - blocks progress thread
let _ = pmixt.lock().unwrap().commit_nb(...);
// ❌ NEVER DO THIS - waits for condition
while !ready { /* spin */ }
// ❌ NEVER DO THIS - modifies shared registry
EVENT_HANDLERS.lock().unwrap().insert(...);
```
```

**Why**: Makes the policy more actionable for contributors.

#### 5. Test - deadlock regression test naming
The test `test_notification_bridge_no_lock_across_user_code` is good but could be more descriptive. Consider renaming to:
`test_event_handler_lock_not_held_during_user_callback`

**Why**: Better communicates the specific property being verified.

### Minor Nitpicks

- In `examples/callback_hop.rs`, ensure the example properly cleans up resources (e.g., deregisters event handlers) before exit to avoid valgrind/ASAN complaints in stricter environments.
- Verify that all new public functions in `threading.rs` have doc examples (even if minimal) to improve discoverability.

### Final Notes

The PR successfully:
1. Introduces zero-cost abstractions (`ProgressContext` ZST)
2. Fixes the lifetime issue in events bridge
3. Maintains lock ordering discipline
4. Provides actionable documentation
5. Includes regression tests

**Ready to merge** with the suggested improvements above (particularly the panic handling in `spawn_from_callback` which addresses a latent safety concern).

**Comment by SedahsDev** [comment]
## Review follow-up applied (`74a5f77`)

Adopted the review suggestions that improve the code without weakening correctness:

| Review item | Disposition |
|---|---|
| `spawn_from_callback` panic handling | **Adopted (corrected).** Catch + branded `eprintln!` + `resume_unwind`. Note: a panic on a Rust worker thread does **not** abort the process by itself (the review’s “default panic hook calls `exit()`” claim is incorrect for non-main threads); we still log for fire-and-forget callers and keep `JoinHandle::join` → `Err`. Silent swallow → `Ok(())` was rejected. |
| Deregistration ordering comment | **Adopted** on the blocking path; expanded with inverse-order / C-failure trade-off. |
| `deregister_event_handler_nb` registry remove | **Bonus fix** found while applying #2: the nb path did not clear `HANDLER_REGISTRY` before the C call (latent UAF window). Now matches the blocking path. |
| `CallbackChannel` sender-drop / `Disconnected` | **Docs + test adopted.** Kept the `Result`/`TryRecvError` API (Empty vs Disconnected) instead of collapsing to `Option` — that distinction is useful. |
| Forbidden-ops concrete examples | **Adopted** on `ProgressContext` docs and `THREADING.md` §6.1. |
| Rename lock-order regression test | **Adopted** → `test_event_handler_lock_not_held_during_user_callback`. |
| Example cleanup / more public doc examples | **Skipped as already satisfied** — `callback_hop.rs` already deregisters; public APIs already have `no_run` examples. |

### Verification
- `cargo test --lib threading::` → **10 passed** (incl. new panic + Disconnected tests)
- `cargo test --lib events::` → **121 passed** (incl. renamed lock-order test)

**Review by SedahsDev — COMMENTED** [review]
## Review Summary

**Verdict: Changes requested.** The registry-based redesign correctly fixes the original use-after-free (the boxed fn is copied out under the lock; a function-pointer copy stays valid even after the Box is dropped), and the lock-discipline, docs, and regression test are good. However, the rewritten bridge breaks the OpenPMIx event-chain completion contract in ways that will hang or crash real event delivery:

1. **The bridge never calls the notification completion `cbfunc` when the registry lookup misses** — the event chain stalls permanently.
2. **The bridge discards PMIx's chain `cbdata` and hands the user handler `NULL`** — a handler that correctly calls its `cbfunc` per the C contract crashes OpenPMIx (`progress_local_event_hdlr` dereferences `notification_cbdata`).
3. **`examples/callback_hop.rs` (the acceptance criterion for #51) never calls `_cbfunc`** — under a real DVM the chain stalls and the example's own blocking `notify_event` hangs forever.

### Verified against OpenPMIx source

- `src/event/pmix_event_notification.c:993-995` — the handler is invoked with `cbfunc=progress_local_event_hdlr` and `cbdata=(void*)chain`; the handler **must** call `cbfunc(status, results, nresults, NULL, NULL, cbdata)` (returning the chain) or the chain never advances to `cycle_events`.
- `src/event/pmix_event_notification.c:705-719` — `progress_local_event_hdlr` casts `notification_cbdata` to `pmix_event_chain_t*` and dereferences it **unconditionally**. Passing `NULL` = segfault.
- Blocking `PMIx_Notify_event` with `cbfunc=NULL` waits on the chain completion (`PMIX_WAIT_THREAD`, line 132-137) → a stalled chain = a permanently blocked notifier.

### What this means in practice

- Registering with `evhdlr = None` (the module docs' own example, `register_event_handler(&codes, &info, None, None)`) now registers the bridge but parks nothing → first matching event stalls the chain.
- `deregister_event_handler` removes the registry entry before the C deregister; an event in flight in that window hits a registry miss → chain stalls instead of completing (the PR's comment calls this "silently dropped" — it is not, it is a hang of every subsequent `notify_event`).
- The deadlock regression test and `test_notification_bridge_unknown_ref_is_noop` both codify the miss ⇒ no-cbfunc behavior as correct.

### Fix direction

- On registry miss, the bridge must still call `cbfunc(status, results, nresults, NULL, NULL, cbdata)` — a pass-through no-op that keeps the chain alive.
- Forward the real `cbdata` (chain) to the user handler instead of `null_mut()`, and document that the handler must return it via `cbfunc`.
- The example must call `_cbfunc` (from the hopped-off app thread is fine — `progress_local_event_hdlr` thread-shifts) after its work completes.

See inline comments for specifics. Additional warnings: `register_event_handler_nb` leaks its state box on synchronous failure (PMIx does *not* invoke the completion callback on the not-initialized / OOM paths — the comment's justification is incorrect); `register_event_handler` silently misbehaves if `cbfunc=Some` is passed; refid `0` is a valid handler id (first registration in a process) which the example's `handler_ref != 0` sentinel mishandles; `spawn_from_callback`'s `catch_unwind`+`resume_unwind` is a no-op that still propagates the panic (and cannot help under `panic=abort` profiles).

*Reviewed by Hermes Agent*

File: src/events.rs [hunk 1]
```diff
@@ -149,13 +181,23 @@ unsafe extern "C" fn notification_bridge(
...
+            .get(&evhdlr_registration_id)
+            .and_then(|b| *b.as_ref())
+    };
+
     // SAFETY: We cast the FFI pointers back to c_void and call the user's
     // callback. The user callback receives opaque pointers and is not expected
     // to dereference them without casting back.
-    unsafe {
-        if let Some(user_fn) = *(cbdata as *mut NotificationFn) {
+    if let Some(user_fn) = user_fn {
```
SedahsDev, line 285: 🔴 **Critical: registry miss never calls the notification completion `cbfunc` → event chain stalls permanently.**

Per the OpenPMIx contract (`src/event/pmix_event_notification.c:993-995`), the handler is invoked with `cbfunc=progress_local_event_hdlr` and `cbdata=(void*)chain`, and the handler **must** call `cbfunc(status, results, nresults, NULL, NULL, cbdata)` to advance the chain to `cycle_events`. If it returns without calling `cbfunc`:

- the chain never completes — `final_cbfunc` never fires,
- any **blocking `notify_event`** (which waits via `PMIX_WAIT_THREAD`, notification.c:132-137) blocks forever,
- any `_nb` notify completion never fires, and the chain object leaks.

This miss path is hit in three real scenarios:

1. **`evhdlr = None` registrations** — `register_event_handler(&codes, &info, None, None)` is the module docs' own example. The bridge is always registered (`Some(notification_bridge)`), but the registry insert is guarded by `if evhdlr.is_some()` — so with `None` nothing is parked, and the *first matching event* stalls the chain.
2. **Deregistration window** — `deregister_event_handler` removes the registry entry *before* `PMIx_Deregister_event_handler`. An event in flight on the progress thread in that window hits a miss here. The PR comment calls this "silently dropped (no UAF)" — it is not a drop, it is a **hang of every subsequent `notify_event`**.
3. **Blocking-register race** — `register_event_handler` inserts the fn *after* the C call returns (line 338), so an event delivered in the gap also misses.

Fix: on miss, call `cbfunc(status, results, nresults, NULL, NULL, cbdata)` — a pass-through no-op that keeps the chain alive. The deadlock regression test and `test_notification_bridge_unknown_ref_is_noop` currently codify the miss ⇒ no-cbfunc behavior; they should be updated to assert `cbfunc` *is* invoked on miss.

src/events.rs
```diff
@@ -149,13 +181,23 @@ unsafe extern "C" fn notification_bridge(
...
+        unsafe {
             user_fn(
```
SedahsDev, line 294: 🔴 **Critical: the chain pointer (`cbdata`) is discarded and `NULL` is handed to the user handler.**

PMIx passes `cbdata = (void*)chain` to the handler, and the handler is required to return that exact pointer via the completion `cbfunc` — `progress_local_event_hdlr` does `pmix_event_chain_t *chain = (pmix_event_chain_t *) notification_cbdata; chain->interim_status = status;` (notification.c:705-719) with **no NULL check**.

So any user handler that correctly follows the C contract (`cbfunc(status, results, nresults, NULL, NULL, cbdata)`) and forwards the `NULL` it was given here will make OpenPMIx dereference `NULL` → **segfault**. The parameter was renamed `_cbdata` to signal "unused" — it must instead be forwarded verbatim to the user fn (and on the registry-miss path, passed through to `cbfunc`), and the `NotificationFn` docs should state that the handler must return that `cbdata` through `cbfunc`.

examples/callback_hop.rs
```diff
@@ -0,0 +1,159 @@
...
+    _cbfunc: pmix::events::pmix_event_notification_cbfunc_fn_t,
+    _cbdata: *mut c_void,
+) {
+    // We are on the progress thread — this marker is documentation, and any
+    // blocking PMIx call here would deadlock progress.
+    let _ctx = pmix::threading::ProgressContext;
+
+    // Hop off: spawn an application thread for the actual handling. We never
+    // join it here (that could block progress and deadlock the process).
+    let _ = pmix::threading::spawn_from_callback(move || {
```
SedahsDev, line 154: 🔴 **Critical: the acceptance-criterion example never calls `_cbfunc` — under a real DVM it deadlocks.**

The handler ignores both `_cbfunc` and `_cbdata`. Per the OpenPMIx event-chain contract, a handler that does not call its completion `cbfunc` stalls the chain: `notify_event` (blocking, used at line 70) will **never return** under `prterun`, and every subsequent event delivery in the job hangs. This example is the deliverable that closes #51, so it should demonstrate the correct pattern:

```rust
unsafe extern "C" fn on_event(..., cbfunc: ..., cbdata: *mut c_void) {
    let _ctx = pmix::threading::ProgressContext;
    let _ = pmix::threading::spawn_from_callback(move || {
        // blocking work …
        if let Some(cbfunc) = cbfunc {
            // SAFETY: complete the event chain from the app thread
            // (progress_local_event_hdlr thread-shifts internally).
            cbfunc(PMIX_SUCCESS, null_mut(), 0, None, null_mut(), cbdata);
        }
    });
}
```

Note `progress_local_event_hdlr` does `PMIX_THREADSHIFT(chain, cycle_events)`, so calling `cbfunc` from the hopped-off app thread is safe — this is the intended hop-off pattern.

src/events.rs
```diff
@@ -321,21 +384,32 @@ pub fn register_event_handler_nb(
...
+        user_cbdata: cbdata,
+    });
 
     // SAFETY: FFI call into PMIx library. The codes slice lives for the
-    // duration of this call. In non-blocking mode, PMIx retains the
-    // evhdlr callback; it must be a static function or otherwise live
-    // for the lifetime of the handler.
+    // duration of this call. The boxed HandlerRegState is reclaimed by
+    // handler_reg_cb_bridge exactly once; on synchronous failure it is
+    // intentionally leaked (PMIx may still invoke the completion callback
```
SedahsDev: ⚠️ **Warning: the leak-on-sync-failure justification is incorrect.**

The comment claims PMIx "may still invoke the completion callback with an error status". Verified against OpenPMIx `PMIx_Register_event_handler` (`src/event/pmix_event_registration.c:900-953`): on the synchronous failure paths (`!initialized` → `PMIX_ERR_INIT`, progress thread stopped → `PMIX_ERR_NOT_AVAILABLE`, `PMIX_NEW`/malloc failure → `PMIX_ERR_NOMEM`) the function returns **before** thread-shifting, and the registration completion callback is **never invoked**. The callback is only guaranteed to fire once the request is thread-shifted (async path).

So on every synchronous failure, `Box<HandlerRegState>` (including the boxed `evhdlr` fn and the user's `cbdata`) leaks permanently. Since PMIx provably cannot call back on those paths, the box can be freed on the `Err` return instead of leaked. At minimum, the comment should state the leak is real and bounded, not claim PMIx may still call back.

src/events.rs
```diff
@@ -242,42 +311,36 @@ pub fn register_event_handler(
...
-
     let status = PmixStatus::from_raw(raw_status);
     if status.is_success() {
-        Ok(raw_status as EventHandlerRef)
+        let handler_ref = raw_status as EventHandlerRef;
+        if evhdlr.is_some() {
+            // Park the user fn for the lifetime of the registration. PMIx may
+            // deliver an event to the bridge the moment registration returns,
+            // so insert before returning to the caller.
+            HANDLER_REGISTRY
```
SedahsDev, line 480: ⚠️ **Race: the registry insert happens *after* the C call returns, not before.**

The comment says "PMIx may deliver an event to the bridge the moment registration returns, so insert before returning to the caller" — but the insert is after `PMIx_Register_event_handler` has returned. Once the C call returns, the C-side handler is live and the progress thread can deliver an event to the bridge before this insert runs, producing a registry miss (see the miss-path comment on line 199) and a stalled chain.

The insert *cannot* move before the call (the refid isn't known until the call returns), so the real fix is the registry-miss path calling `cbfunc` as a pass-through — then this window becomes a harmless dropped event instead of a hang. Worth a code comment acknowledging the window.

examples/callback_hop.rs
```diff
@@ -0,0 +1,159 @@
...
+            }
+            Err(_) => {
+                println!("[app thread] no get_nb completion within 5s (standalone run?)");
+                break;
+            }
+        }
+    }
+
+    // ── 5. Cleanup ─────────────────────────────────────────────────────────
+    if handler_ref != 0 {
```
SedahsDev: ⚠️ **`handler_ref != 0` is not a valid failure sentinel.**

OpenPMIx assigns handler indices from `pmix_globals.events.nhdlrs`, which starts at `0` (`src/event/pmix_event.h:145`, `pmix_event_registration.c:662-664`) — so the **first handler registered in a process legitimately gets refid `0`**. `register_event_handler` returns `Ok(0)` for it (a positive/non-error status per `PmixStatus::is_success`), and this example would then skip deregistration, leaking the registry entry and the C handler. Track failure via the `Result` instead:

```rust
let Ok(handler_ref) = pmix::events::register_event_handler(&[], &info, Some(on_event), None) else {
    println!("register_event_handler failed (no DVM?): …");
    return; // or continue without events
};
```

src/threading.rs
```diff
@@ -0,0 +1,403 @@
...
+    F: FnOnce() + Send + 'static,
+{
+    std::thread::Builder::new()
+        .name("pmix-callback-hop".to_string())
+        .spawn(move || {
+            // Isolate panics to this hop thread. Log a clear diagnostic for
+            // operators (fire-and-forget callers never join), then resume so
+            // `JoinHandle::join` still surfaces the payload instead of a
+            // silent `Ok(())`.
+            if let Err(payload) = std::panic::catch_unwind(std::panic::AssertUnwindSafe(f)) {
```
SedahsDev, line 161: ⚠️ **`catch_unwind` + `resume_unwind` is a no-op that still propagates the panic.**

Catching the panic, logging, then immediately `resume_unwind` re-raises it: the hop thread still panics (default panic hook prints a second, duplicate message), and under `panic = "abort"` profiles (common in HPC release builds) `catch_unwind` cannot intercept at all — the process aborts regardless. So the doc claim "a panic inside `f` is **isolated** to the hop thread (it does not abort the process by itself)" is only true in the same sense as any spawned thread without this wrapper.

Decide what the helper is for:
- **True isolation** (the earlier review suggestion): log and *swallow* — do not `resume_unwind`; the joined handle then reports `Ok(())`, which must be documented.
- **Surfacing to joiners**: drop the `catch_unwind` entirely and let the panic propagate — identical behavior minus the extra `eprintln!`.

Also note `eprintln!` in a library is a side channel; a `log`-based or feature-gated diagnostic would be more appropriate.

SedahsDev:
## Review follow-up applied (`26bdc11`)

Adopted the second-pass review items that fix real hang/crash paths in the events bridge (and the related example / docs fixes). First-pass items were already on the branch at `74a5f77`.

| Review item | Disposition |
|---|---|
| Registry miss never calls notification `cbfunc` (chain stall) | **Adopted.** `notification_bridge` pass-through-calls `cbfunc(status, results, nresults, None, null, cbdata)` on miss. |
| Chain `cbdata` discarded / `NULL` handed to user handler | **Adopted.** `cbdata` forwarded verbatim to the user fn (and on the miss path). `NotificationFn` docs state the completion contract. |
| `callback_hop` never calls `_cbfunc` (DVM deadlock) | **Adopted.** Handler hops off, then completes the chain from the app thread (`progress_local_event_hdlr` thread-shifts). |
| `register_event_handler_nb` leak on sync failure + wrong comment | **Adopted.** Free `HandlerRegState` on `Err`; comment cites OpenPMIx sync-failure paths that never invoke the completion callback. |
| Blocking register insert race after C return | **Adopted (comment).** Insert cannot move earlier (refid unknown); miss path now pass-through-completes so the window is a dropped delivery, not a hang. |
| `handler_ref != 0` failure sentinel | **Adopted.** Example tracks registration via `Option`/`Result`; docs note OpenPMIx first refid is `0`. |
| `register_event_handler(..., Some(cbfunc))` silent misbehavior | **Adopted.** Blocking API rejects non-`None` `cbfunc` with `PMIX_ERR_BAD_PARAM`; route async through `register_event_handler_nb`. |
| `catch_unwind` + `resume_unwind` is a no-op / false isolation claim | **Adopted (docs corrected).** Kept log + `resume_unwind` so joiners still see `Err`; docs no longer claim true isolation under `panic = "abort"`. Silent swallow rejected (would hide panics as `Ok(())`). |
| Deregistration “silently dropped” wording | **Adopted.** Comment now says miss → pass-through completion (no hang, no UAF). |

### Verification
- `cargo test --lib threading::` → **10 passed**
- `cargo test --lib events::` → **122 passed** (incl. `test_notification_bridge_unknown_ref_invokes_cbfunc`, cbdata-forward check, BAD_PARAM on blocking+cbfunc)
- `cargo build --example callback_hop` → ok

src/events.rs
```diff
@@ -151,11 +214,21 @@ unsafe extern "C" fn notification_bridge(
...
+            .get(&evhdlr_registration_id)
+            .and_then(|b| *b.as_ref())
+    };
+
+    if let Some(user_fn) = user_fn {
+        // SAFETY: cast FFI pointers to c_void for the user-facing NotificationFn
+        // ABI. Forward OpenPMIx's chain pointer (`cbdata`) and completion
+        // `cbfunc` verbatim — the handler must return them via cbfunc.
+        unsafe {
             user_fn(
```
SedahsDev, line 294: ⚠️ **Panic across FFI + event-chain stall on the hit path.** `user_fn(...)` runs inside this `unsafe extern "C"` bridge on the PMIx progress thread. If the user's handler panics: (1) the unwind crosses the C boundary (`notification_bridge` was called from OpenPMIx), which is UB; and (2) `cbfunc` is never called, so the OpenPMIx event chain stalls — a blocking `notify_event` hangs forever, the exact failure mode this PR exists to prevent. The registry-miss path below correctly completes the chain; the hit path should too.

Suggestion: wrap the call in `std::panic::catch_unwind` (the same care `spawn_from_callback` already takes), and on panic pass-through-complete the chain via `cbfunc(status, results, nresults, None, null_mut(), cbdata)` before logging/resuming the panic. Note the crate-wide `.lock().expect("mutex poisoned ...")` pattern is also a latent panic source on the progress thread — `unwrap_or_else(PoisonError::into_inner)` would be cheap hardening, though poisoning is practically impossible today since user code never runs under these locks.

src/events.rs
```diff
@@ -242,42 +359,40 @@ pub fn register_event_handler(
...
-    }
-
     let status = PmixStatus::from_raw(raw_status);
     if status.is_success() {
-        Ok(raw_status as EventHandlerRef)
+        let handler_ref = raw_status as EventHandlerRef;
+        if evhdlr.is_some() {
+            // Park the user fn for the lifetime of the registration.
+            //
+            // Race window: the C-side handler is live as soon as
```
SedahsDev, line 475: ⚠️ **Documented race window has a real consequence: a silently *dropped* event, not just a non-hang.** An event delivered between the C-side handler becoming live and this insert is pass-through-completed with no user callback — the user's handler never sees it. For loss-sensitive notifications (e.g. `PMIX_EVENT_JOB_END` in `examples/callback_hop.rs`) that is a missed event, not merely 'no hang'. I agree it's inherent to the API (the refid is only known after the call returns) and the pass-through is the right degradation — but please surface this limitation in the public `register_event_handler` docs (currently only the internal comment) so users registering for critical events know a delivery can be dropped in the registration window.

src/events.rs
```diff
@@ -131,15 +160,49 @@ pub type HandlerRegCbFn =
...
+///
+/// PMIx retains an event handler — and its `cbdata` — for the lifetime of the
+/// registration, so the user's [`NotificationFn`] must outlive it. We store
+/// the function here, keyed by the reference ID that PMIx passes to the
+/// notification bridge as its first argument, and free it at deregistration.
+/// This follows the bridge policy in [`crate::threading`]: the registry lock
+/// is held only to copy the function pointer out, never across the user call,
+/// and no callback data rides in a raw `cbdata` pointer.
+type HandlerRegistry = HashMap<EventHandlerRef, Box<NotificationFn>>;
+static HANDLER_REGISTRY: LazyLock<Mutex<HandlerRegistry>> =
```
SedahsDev, line 195: 💡 **Registry entries are only freed at deregistration.** A user who registers a handler and never deregisters (or whose process takes the `finalize` path without deregistering) leaks the `Box<NotificationFn>` for the process lifetime. Since the crate already documents a no-reinit policy, clearing `HANDLER_REGISTRY` in the finalize/disconnect path would be a safe, complete cleanup. Low severity — bounded by the number of registrations — but cheap to fix.

SedahsDev (commented):
## Hermes Agent review — verified against OpenPMIx source, build, and tests

**Verdict: Comment / approve-with-suggestions** — the core fix is correct and I verified it end-to-end. Nothing blocking; 3 inline points, 2 are worth taking before merge (panic-across-FFI on the hit path; documenting the dropped-delivery window), 1 is a minor cleanup.

### ✅ Verified correct (grounded in `/home/bzf/openpmix/src/event/`)
- **Chain-pointer fix**: forwarding `cbdata` verbatim (instead of `null_mut()`) is the real bug fix — `progress_local_event_hdlr` does `chain = (pmix_event_chain_t*)notification_cbdata` with no NULL check, so the old code was a guaranteed segfault/stall. The new pass-through on registry miss exactly matches the required `cbfunc(status, results, nresults, NULL, NULL, cbdata)` contract.
- **Registry replaces boxed cbdata**: correct UAF fix. Safe in part because `NotificationFn` is a plain fn pointer (code address) — copy-out-under-lock then free is not a UAF. Worth noting that invariant in a comment, since the analysis breaks if the type ever becomes a capturing closure.
- **Ref IDs are 0-based**: the `Ok(0)` is a valid first refid documentation is correct (verified `pmix_event_registration.c`: counter assigned before increment).
- **`register_event_handler_nb` box lifecycle**: exactly-once reclaim via the completion bridge, and freed on synchronous `Err` — correct, since sync-failure paths return before thread-shift and never invoke the completion callback. The old 'intentionally leaked' comment was wrong; this PR fixes the leak too.
- **Deregistration ordering**: remove-before-C-call is safe; the miss window is covered by the pass-through. Also confirmed the audit claim: `data_ops`/`query_log` scope registry locks the same way (user code outside the lock).

### 🧪 Tested locally (PMIx 5.0.7 system install, committed bindings preserved)
- `cargo build --examples` ✅ · `cargo test --lib`: **1757 passed, 0 failed** (10 threading + 122 events incl. the 8 new bridge/registry tests and the deadlock regression test) ✅ · PR doctests (threading module, `spawn_from_callback`, `CallbackChannel`) pass ✅ · `examples/callback_hop` runs standalone with graceful no-DVM exit ✅ · clippy: no new warnings on PR files (the one `unneeded late initialization` at lib.rs:3711 and the failing `PmixClient` doctest at lib.rs:3500 are both pre-existing on `main`, untouched by this PR — as is the `PMIx_Progress_thread_stop` bindgen issue on PMIx 5.0.7, unrelated to these changes).

### 📝 Nits
- PR description names the deadlock regression test `test_notification_bridge_no_lock_across_user_code`; the actual name is `test_event_handler_lock_not_held_during_user_callback` — align the description.
- `handler_reg_cb_bridge` returns early on NULL `cbdata` without firing the user's completion callback (defensive only — PMIx always passes our box back; no action needed, just noting the user's `cbfunc` would never fire in that hypothetical path).

### 🔗 Inline comments
1. `src/events.rs:232` — panic across FFI / chain stall on the registry-hit path (suggest `catch_unwind` + pass-through completion).
2. `src/events.rs:385` — the registration-window race *drops* events; document it in the public API docs.
3. `src/events.rs:177` — registry never cleared at finalize; consider cleanup in the disconnect path.

Overall: this is a solid, well-documented fix for a genuine UAF/stall bug. Ship it after considering #1 and #2.

---
*Reviewed by Hermes Agent*

SedahsDev:
## Review follow-up applied (`aca7840`)

Adopted the third-pass review items that improve progress-thread safety and finalize hygiene. Pass-1 (`74a5f77`) and pass-2 (`26bdc11`) items remain on the branch.

| Review item | Disposition |
|---|---|
| Panic across FFI + chain stall on registry-hit path | **Adopted (corrected).** `notification_bridge` wraps the user call in `catch_unwind`. On panic: log, pass-through-complete the OpenPMIx chain via `cbfunc`, and **contain** the panic (no `resume_unwind` into C — that would be UB). |
| `NotificationFn` was `extern "C"` (`nounwind`) so panics abort before catch | **Bonus fix** required for the item above to work. `NotificationFn` is now a Rust-ABI `unsafe fn`; only `notification_bridge` stays `extern "C"` as the sole C→Rust boundary. Docs state the ABI choice. |
| Registration-window race drops events (document publicly) | **Adopted.** Public `# Registration-window delivery` section on `register_event_handler` explains the inherent post-return insert gap and that loss-sensitive codes can miss a racing delivery. |
| `HANDLER_REGISTRY` never cleared at finalize | **Adopted.** `events::clear_handler_registry()` called from client `PmixClient::disconnect`, tool `PmixTool::disconnect`, and server `PmixServer::disconnect` after the C finalize call. |
| Poisoned registry mutex on progress thread | **Adopted (light).** Bridge lookup uses `unwrap_or_else(PoisonError::into_inner)` instead of `expect`. |
| PR description deadlock-test name mismatch / null-cbdata early return note | **Skipped.** Naming already matches live code from pass-1; null-cbdata path is defensive-only as the review noted — no action. |

### Verification (ad-hoc / focused; PMIx 5.0.7 host with committed bindings)
- `cargo test --lib notification_bridge` → **5 passed** (incl. `test_notification_bridge_user_panic_completes_chain_and_is_contained`)
- `cargo test --lib clear_handler_registry` → **1 passed**
- `cargo test --lib events::` → **124 passed**
- `cargo test --lib threading::` → **10 passed**
- `cargo build --example callback_hop` → ok

PR #369, opened 2026-08-06
test(threading): multi-thread + external-progress integration (#54)

### LINKED ISSUE #54: threading: multi-thread + external-progress integration tests
### Goals
1. N threads: clone `PmixClient`, concurrent put (distinct keys) + fence under `prterun`/daemon harness
2. Concurrent `_nb` completions
3. `external_progress=true` + host thread `progress()` while workers issue ops
4. Callback must-not-block regression (timeout) — ties to #51
5. Optional TSAN job (nightly)

### Acceptance
- [ ] At least (1) and (3) automated or scripted (`tests/` + `scripts/run_daemon_tests.sh`)
- [ ] Documented how to run
- [ ] Uses **process-wide** `PmixClient` (not bare `Context` / not per-thread init)

### Depends on
#48 done; #46/#47 done (InitOptions + progress_thread_stop). #51 for (4).

SedahsDev:
## Summary
- Adds `tests/threading_mt_via_prterun.rs` covering issue #54 goals (1)–(4) with **process-wide** `PmixClient` (no bare `Context`, no per-thread init).
- Wires `./scripts/run_daemon_tests.sh THREADING` (build once, run each ignored case under its own `prterun -np 1`).
- Documents how to run in `THREADING.md` + `README.md`.
- Companion example: `examples/external_progress_mt.rs`.
- **Bugfix:** `InfoBuilder` now encodes `PMIX_BOOL` attributes (`pmix.evext`, etc.) as real `u8` flags instead of the C strings `"1"`/`"0"`, so `InitOptions::external_progress(true)` is actually honored by OpenPMIx.

## Test plan
- [x] `cargo test --test threading_mt_via_prterun -- --test-threads=1` (standalone)
- [x] `./scripts/run_daemon_tests.sh THREADING` under PRTE 4.1 + OpenPMIx 6.1
- [x] All four DVM cases pass: put/fence, fence_nb, callback timeout, external_progress
- [x] `cargo test --lib test_infobuilder_ -- --test-threads=1`

Closes #54

### DIFF
--- a/scripts/run_daemon_tests.sh
+++ b/scripts/run_daemon_tests.sh
@@ -1,71 +1,134 @@
 #!/bin/bash
-# Run pmix-rs tests that require a PMIx daemon.
-# Reads the PRTE URI from the systemd-managed service and exports PMIX_SERVER_URI.
+# Run pmix-rs tests that require a PMIx daemon / DVM.
 #
 # Usage:
-#   ./scripts/run_daemon_tests.sh          # Run all daemon tests
-#   ./scripts/run_daemon_tests.sh TOOL     # Run only tool_tool_init tests
-#   ./scripts/run_daemon_tests.sh LIB      # Run only lib_core_api daemon tests
-#   ./scripts/run_daemon_tests.sh FABRIC   # Run only fabric daemon tests
-#   ./scripts/run_daemon_tests.sh COV      # Run coverage with daemon tests included
+#   ./scripts/run_daemon_tests.sh              # Run all daemon tests
+#   ./scripts/run_daemon_tests.sh TOOL         # tool_tool_init tests
+#   ./scripts/run_daemon_tests.sh LIB          # lib_core_api daemon tests
+#   ./scripts/run_daemon_tests.sh FABRIC       # fabric daemon tests
+#   ./scripts/run_daemon_tests.sh THREADING    # multi-thread + external-progress (#54)
+#   ./scripts/run_daemon_tests.sh COV          # coverage with daemon tests included
 #
-# Prerequisites:
-#   systemctl --user start prte
+# Prerequisites (TOOL/LIB/FABRIC/ALL — tool URI path):
+#   systemctl --user start prte   # or a PRTE system-server writing $URI_FILE
+#
+# Prerequisites (THREADING — DVM / prterun path):
+#   OpenPMIx ≥ 6.1 and PRTE ≥ 4.1 on PATH (same libpmix), e.g.:
+#     export PMIX_PREFIX=$HOME/.local/openpmix-6.1.0
+#     export PATH=/path/to/prte-4.1/bin:$PATH
+#     export LD_LIBRARY_PATH=$PMIX_PREFIX/lib:$LD_LIBRARY_PATH
+#   prterun must work: `prterun --version`
 
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 PROJECT_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
-URI_FILE="/run/user/1000/prte/uri"
+URI_FILE="${PMIX_TEST_URI_FILE:-/run/user/$(id -u)/prte/uri}"
 
-# Check PRTE is running
-if ! systemctl --user is-active prte &>/dev/null; then
-    echo "ERROR: prte systemd service is not running."
-    echo "Start it with: systemctl --user start prte"
-    exit 1
-fi
+require_uri_daemon() {
+    if [ ! -f "$URI_FILE" ]; then
+        # Best-effort: systemd unit may not exist on all hosts.
+        if systemctl --user is-active prte &>/dev/null; then
+            :
+        elif systemctl --user list-unit-files 2>/dev/null | grep -q '^prte'; then
+            echo "ERROR: prte systemd service is not running."
+            echo "Start it with: systemctl --user start prte"
+            exit 1
+        else
+            echo "ERROR: PRTE URI file not found at $URI_FILE"
+            echo "Start a PRTE system-server that writes the URI, or set PMIX_TEST_URI_FILE."
+            exit 1
+        fi
+    fi
+    if [ ! -f "$URI_FILE" ]; then
+        echo "ERROR: PRTE URI file not found at $URI_FILE"
+        exit 1
+    fi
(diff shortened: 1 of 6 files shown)


SedahsReview (dismissed):
Please add panic catching in all callback bridges (e.g., in allocation_callback_bridge and similar functions in other modules) to avoid undefined behavior when unwinding through C ABI.

scripts/run_daemon_tests.sh
```diff
@@ -1,71 +1,134 @@
...
-        cargo test --test fabric_fabric_comprehensive -- --ignored
+        cargo test --test fabric_fabric_comprehensive -- --ignored --test-threads=1
+        ;;
+    THREADING)
+        require_prterun
+        echo "Running multi-thread + external-progress tests (issue #54)..."
+        # Build once, then run the test binary under prterun (avoids nested cargo
+        # and lets us bound each case with timeout).
+        cargo test --test threading_mt_via_prterun --no-run
+        BIN=$(ls -1t target/debug/deps/threading_mt_via_prterun-* | grep -v '\.d$' | head -1)
```
SedahsReview: 💡 **Suggestion (non-blocking):** under `set -euo pipefail`, if the glob `target/debug/deps/threading_mt_via_prterun-*` matches nothing, `ls` fails and the command-substitution failure terminates the script *before* the friendly `could not locate` error on line 88 — so that branch is dead code. Also, `ls -1t | head -1` picks the newest of possibly several build variants (different hashes). `cargo test --no-run --message-format=json` (parse `executable` from the `compiler-artifact` messages) gives the exact binary path deterministically. Minor — the 60s `timeout` per case is a nice fail-closed touch.

tests/threading_mt_via_prterun.rs
```diff
@@ -0,0 +1,462 @@
...
+    let client2 = client.clone();
+    let proc2 = proc.clone();
+    let (tx, rx) = std::sync::mpsc::channel();
+    thread::spawn(move || {
+        let _ = client2.is_live();
+        let r = pmix::fence(&proc2, None);
+        let _ = tx.send(r);
+    });
+
+    match rx.recv_timeout(Duration::from_secs(8)) {
```
SedahsReview, line 453: 💡 **Suggestion (non-blocking):** this regression test never asserts — both the completed-fence and the timed-out-fence branches are accepted, and nothing records that `BlockingFenceCb::on_complete` actually ran. A future change that stopped invoking callbacks (or never submitted the `fence_nb`) would still pass. Consider setting an `AtomicBool` at the end of `on_complete` (after the condvar wait) and asserting it after `recv_timeout` returns, plus logging which branch fired. Boundedness is good — it's currently a hang-smoke-test rather than a behavior assertion.

SedahsReview (approved):
## Hermes Agent Review — Approve ✅

Reviewed the full PR diff (single commit `1692f52` on base `3e2462a`; 6 files, +714/−65), plus local compile and test runs.

### The bugfix is real and correct
`InfoBuilder` now encodes `PMIX_BOOL` attributes as scalar `u8` flags instead of C strings. Verified against the installed OpenPMIx 6.1 headers (`pmix_common.h:1762`: `PMIX_BOOL ... converted to/from native true/false to uint8 for pack/unpack`): the old `add_string_key(..., "1"/"0", PMIX_BOOL)` made PMIx read the first byte of the string — `'0'` = 0x30 ≠ 0 — so `external_progress(false)`, `bind_required(false)` and `progress_thread_flush(false)` were all silently passed as **true**. The new `InfoEntryBool` (owned `u8` storage) encodes 0/1 correctly, and it also fixes a latent dangling-pointer UB in the old `collect_data` (pointer to a stack `bool` stored past its lifetime). `build()` counts all three entry vectors, keeps `idx` consistent across the loops, and frees via `PMIx_Info_free` on load errors. The new `init_options_external_progress_builds` test dereferences the built `Info` and asserts key/type/flag — real validation of the encoding.

### Tests & harness
- Clean process-wide `PmixClient` model, per-case `prterun` isolation for the stateful `external_progress` test, `is_dvm_launched` guards, bounded waits (5s/8s/30s) and a 60s harness `timeout` so hangs fail closed.
- The `atexit` finalize safety net matches the documented `daemon_helper` pattern (prterun aborts DVM processes that exit without finalize).
- CI green: Regular tests, Clippy, Daemon tests, **DVM tests (via prterun)** — all success. Locally: `--no-run` build clean, the 4 standalone tests pass, and the 9 `test_infobuilder_*` lib tests pass.

### Suggestions (non-blocking, see inline)
1. `run_daemon_tests.sh:86` — the `ls <glob> | head -1` binary lookup makes the friendly `could not locate` branch dead code under `set -euo pipefail`; `--message-format=json` is more deterministic.
2. `callback_must_not_block_progress_timeout` never asserts anything — neither branch fails and nothing proves the blocking callback actually ran; an `AtomicBool` in `on_complete` would make it a real regression guard.

**Verdict: Approve.** The InfoBuilder fix is a genuine, verified bugfix, and the test coverage for all four issue #54 goals is thoughtful and fail-closed.

SedahsDev:
## Review follow-up applied (`f8fb614`)

| Review item | Disposition |
|---|---|
| `run_daemon_tests.sh` `ls <glob> \| head -1` binary lookup (dead friendly error under `set -e`) | **Adopted.** Resolve executable via `cargo test --no-run --message-format=json` + `compiler-artifact.executable`; keep explicit missing/non-executable errors. |
| `callback_must_not_block_progress_timeout` never asserts callback ran | **Adopted.** `BlockingFenceCb` sets an `AtomicBool` after the condvar wait; test waits briefly then asserts it, and logs which secondary-fence branch fired. |
| Dismissed review: panics in all callback bridges | **Rejected (out of scope).** Already dismissed; not part of this PR’s diff. Bridge panic policy belongs in a dedicated events/hop follow-up, not this MT integration test PR. |

### Verification (ad-hoc/focused)
- `cargo test --test threading_mt_via_prterun -- --test-threads=1` → **4 passed**, 4 ignored (DVM)
- JSON binary resolve smoke → executable path OK
- `cargo test --lib test_infobuilder_ -- --test-threads=1` → **9 passed**
- DVM/`prterun` cases **not** re-run here (`prterun` not on PATH on this host); CI DVM job remains the gate

Prior verdict was **Approve** with green CI on the original commit; this is a small non-blocking follow-up on the same head branch.

PR #371, opened 2026-08-07
test(threading): add public Send and Sync matrix

### LINKED ISSUE #66: threading: static_assertions matrix for all public handle types
## Compile-time Send/Sync matrix

Today only:
- `Info: !Send + !Sync`
- `PmixClient: Clone + Send + Sync`

### Goals
Central test (or per-module) asserting intended auto-traits for every public type in THREADING inventory:
- Sessions: `PmixClient` (+ future server/tool) → `Send + Sync`
- POD: `Proc`, status/enums → `Send + Sync`
- C-owned: `Info`, buffers, fabric, results → `!Send + !Sync` (as each lands in #50)

### Acceptance
- [ ] One place to extend when new handles ship
- [ ] CI fails if a raw-ptr wrapper silently becomes `Send`

### Depends on
#50 for completing !Send marks; can start with known types now.

SedahsDev:
## Summary
- Add a centralized compile-time `Send`/`Sync` classification matrix for the public PMIx API.
- Document ownership, callback, session, identity, and thread-affinity rules in `THREADING.md`.
- Keep server and tool `PmixServerHandle` aliases clearly distinguished.

## Validation
- `cargo test --lib --no-default-features` — 1,753 passed, 0 failed, 29 ignored.
- `cargo check --all-targets --no-default-features` — passed.
- `cargo check --tests --no-default-features` — passed.
- `rustfmt --check src/threading_assert.rs` — passed.
- `git diff --check` — passed.

The full all-target test command reaches the existing `data_serialization_bench` target, which requires PMIx initialization for `PMIx_Data_load`; on an uninitialized process it returns `ErrInit` and the benchmark currently unwraps it. This is an existing benchmark/environment limitation, not a failure in the changed files.

Closes #66

Pushed commit: `68b3acf79ddd275eca74ad48c44d9d153bb2b8e0`

### DIFF
--- a/src/threading_assert.rs
+++ b/src/threading_assert.rs
@@ -1,69 +1,154 @@
-//! Compile-time `Send`/`Sync` matrix for public session vs C-owned handle types.
+//! Compile-time `Send`/`Sync` matrix for the public PMIx API.
 //!
-//! See [THREADING.md](../THREADING.md) and issue #50.
+//! This is intentionally centralized: when a public wrapper is added or its
+//! ownership model changes, add its intended auto-traits here.  The negative
+//! assertions are as important as the positive ones.  In particular, they
+//! prevent a raw-pointer-bearing wrapper from silently becoming movable or
+//! shareable after a representation change.
+//!
+//! See [THREADING.md](../THREADING.md), issue #50, and issue #66.
 
 #[cfg(test)]
 mod asserts {
     use static_assertions::{assert_impl_all, assert_not_impl_any};
 
-    // ── Sessions: must stay shareable ─────────────────────────────────────
-    use crate::{PmixClient, PmixClientState, Proc};
-    use crate::server::{PmixServer, PmixServerState};
-    use crate::tool::{PmixTool, PmixToolState};
+    // ── Sessions, identities, POD, and scalar values ──────────────────────
+    //
+    // These values contain either process-safe session state, copied process
+    // identity, scalar data, or Rust-owned memory.  They are deliberately
+    // usable from worker threads.
+    assert_impl_all!(crate::PmixClient: Clone, Send, Sync);
+    assert_impl_all!(crate::PmixClientState: Send, Sync);
+    assert_impl_all!(crate::server::PmixServer: Clone, Send, Sync);
+    assert_impl_all!(crate::server::PmixServerHandle: Clone, Send, Sync);
+    assert_impl_all!(crate::server::PmixServerState: Send, Sync);
+    assert_impl_all!(crate::tool::PmixTool: Clone, Send, Sync);
+    assert_impl_all!(crate::tool::PmixToolState: Send, Sync);
+    assert_impl_all!(crate::tool::PmixToolHandle: Clone, Send, Sync);
+    assert_impl_all!(crate::tool::PmixServerHandle: Clone, Send, Sync);
+    assert_impl_all!(crate::Proc: Clone, Send, Sync);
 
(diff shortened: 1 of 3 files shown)


src/threading_assert.rs
```diff
@@ -1,69 +1,159 @@
...
+    //
+    // A function pointer has no captured state.  Callback wrapper structs are
+    // movable because their trait objects require `Send`, but they are not
+    // advertised as concurrently shareable (`Sync` is intentionally absent).
+    assert_impl_all!(crate::events::EventHandlerRef: Send, Sync);
+    assert_impl_all!(crate::events::NotificationFn: Send, Sync);
+    assert_impl_all!(crate::events::HandlerRegCbFn: Send, Sync);
+    assert_impl_all!(crate::events::OpCbFn: Send, Sync);
+    assert_impl_all!(crate::events::pmix_event_notification_cbfunc_fn_t: Send, Sync);
+    assert_impl_all!(crate::groups::pmix_group_opt_t: Send, Sync);
```
SedahsReview: 💡 **Suggestion (non-blocking):** `pmix_group_opt_t` is not a function pointer — in OpenPMIx 6.1 it's a scalar enum (`pmix_common.h:1934-1937`: `typedef enum { PMIX_GROUP_DECLINE, PMIX_GROUP_ACCEPT } pmix_group_opt_t;`). The `Send + Sync` assertion is correct, but the enclosing comment "A function pointer has no captured state" mischaracterizes it. Consider moving it up to the scalar/POD section (e.g. next to the other PMIx enum mirrors), or widening the section comment to "Function pointers, handles, and callback carriers" — note `EventHandlerRef` is a `usize` registry handle rather than a fn pointer either.

SedahsReview (approved):
## Review: Approved ✅

Reviewed the full diff against base `5438d85` and verified the matrix by compiling and running it locally (OpenPMIx 6.1 headers via bindgen).

### Verification performed
- `cargo check --tests --no-default-features` and `cargo check --all-targets --no-default-features` — pass; every positive and negative `Send`/`Sync` assertion in `threading_assert.rs` holds at compile time.
- `cargo test --lib --no-default-features` — **1,753 passed, 0 failed, 29 ignored**, matching the PR body exactly.
- Cross-checked each doc claim against the source: `server::PmixServerHandle = PmixServer` alias (`server/session.rs:226`), distinct `tool::PmixServerHandle { proc: Proc }` (`tool.rs:430`), `PmixEnvar` (CString fields), `PmixCredential` (Rust `Vec<u8>`), fn-pointer aliases (`Option<unsafe extern "C" fn>`).
- `PmixServerModule::as_c_ptr` indeed casts `self as *const Self as *const ffi::pmix_server_module_t` — the documented reason for excluding it from the matrix is accurate.
- Negative assertions cover the full 5.2 table including the newly added `PmixPayload`, `PmixValueBuilder`, `PmixPdata`, and `CredentialResults`.

### Non-blocking suggestions (1)
- `src/threading_assert.rs:82` — `pmix_group_opt_t` is placed under the "function pointer" comment but is a scalar enum; see inline comment.

### Notes
- The matrix is `#[cfg(test)]`-gated, so the compile-time guarantees apply to test builds; that matches the PR's `test(threading)` scope and the pre-existing design — fine as-is, just worth remembering the guarantee does not extend to downstream consumers' builds.
- No prior review rounds on this PR; the two `docs(threading)` follow-up commits are internal polish and look good.

Thanks for the thorough documentation — the `Arc<Mutex<T>>` clarification and the `PmixByteObject` disambiguation in THREADING.md are genuinely useful corrections.

PR #393, opened 2026-08-08
fix(events): close first-event drop window in blocking register_event_handler

### LINKED ISSUE #391: bug: blocking register_event_handler has a first-event drop window between refid return and registry insert
## Summary

In the blocking path, `register_event_handler` (`src/events.rs:424-510`) calls `PMIx_Register_event_handler` synchronously, gets the refid, and then inserts the user fn into `HANDLER_REGISTRY`. An event matching the new registration that fires in the window between PMIx returning the refid and the registry insert will hit `notification_bridge` as a registry miss and be **pass-through-completed** — the event is silently dropped at the Rust layer (the code documents this at `src/events.rs:254-258` and `:323-331`).

## Impact

- A process that registers a handler for a frequently-firing event class (e.g. `ErrJobAborted`, IOF, heartbeat) can lose the first event(s) if the registration races with an in-flight delivery. Blocking `notify_event` won't hang (the chain is completed), but the notification is lost.
- For fault-detection handlers this is exactly the event you cannot afford to miss.

## Suggested fix

- Buffer pre-refid deliveries: since the refid is unknown until the C call returns, register the user fn under a provisional key (or a pending queue keyed by "no refid yet") and re-key on insert; the bridge can consult both maps. Or
- Have the bridge, on a registry miss with a `pending_registrations` entry, deliver to the pending fn; or
- Document the window as a hard limitation and recommend the `_nb` registration path for delivery-critical handlers.

Severity is low (narrow window, documented), but the drop is silent and affects correctness for event-driven clients.

SedahsDev:
Closes #391

## Summary
- Park blocking event handlers in a provisional registry before entering PMIx.
- Resolve matching provisional handlers during the refid-return window, then re-key them under the returned refid.
- Preserve pass-through completion for genuine registry misses and clear pending handlers during finalize cleanup.

## Mechanism
This uses a pending queue keyed by the registered event codes. The blocking path parks the handler before the synchronous C call and removes/re-keys it after PMIx returns; the notification bridge consults the normal registry first and then matching pending registrations.

## Module placement rationale
The pending state is kept beside HANDLER_REGISTRY in src/events.rs, where the bridge, blocking registration, deregistration, and finalize cleanup already coordinate handler lifetime and chain completion. No public API signatures changed.

## Test plan
- Local mock-based events unit tests, including pending delivery/re-keying and existing pass-through, panic, and deregistration behavior.
- cargo build
- cargo test (daemon/DVM tests are unavailable locally and remain CI-only)
- cargo fmt --check
- cargo clippy --all-targets -- -D warnings (currently reports pre-existing repository lint failures outside this change)

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -274,7 +274,9 @@ unsafe fn complete_event_chain(
 /// bridge still calls `cbfunc` as a pass-through. During blocking registration,
 /// a matching provisional handler is used while its refid is being returned by
 /// PMIx. A delivery for a refid currently being deregistered is passed through
-/// rather than matched against a provisional registration.
+/// rather than matched against a provisional registration. The provisional
+/// match is only for the registration's matching event codes; a foreign refid
+/// must not be attributed to an unrelated default handler.
 ///
 /// User-handler panics are caught: unwinding across the C→Rust FFI boundary
 /// is undefined behaviour, and a panic that skips `cbfunc` would stall the
@@ -308,15 +310,29 @@ unsafe extern "C" fn notification_bridge(
         {
             return None;
         }
-        let pending = PENDING_REGISTRATIONS
-            .lock()
-            .unwrap_or_else(|e| e.into_inner());
-        pending
-            .iter()
-            .find(|registration| {
-                registration.codes.is_empty() || registration.codes.contains(&status)
-            })
-            .and_then(|registration| *registration.user_fn.as_ref())
+        let pending_fn = {
+            let pending = PENDING_REGISTRATIONS
+                .lock()
+                .unwrap_or_else(|e| e.into_inner());
+            pending
+                .iter()
+                .find(|registration| {
+                    registration.codes.is_empty() || registration.codes.contains(&status)
+                })
+                .and_then(|registration| *registration.user_fn.as_ref())
+        };
+        pending_fn.or_else(|| {
+            // The re-key holds PENDING_REGISTRATIONS while inserting into the
+            // registry before popping. If the pending consult misses, that
+            // critical section may have completed between the first registry
+            // read and this consult, so re-read the registry. Each lock is
+            // scoped and released before the next, preserving no-ABBA order;
+            // this only runs on the registry-miss plus pending-miss path.
+            let registry = HANDLER_REGISTRY.lock().unwrap_or_else(|e| e.into_inner());
+            registry
+                .get(&evhdlr_registration_id)
+                .and_then(|b| *b.as_ref())
+        })
     });
 
     if let Some(user_fn) = user_fn {
@@ -459,6 +475,7 @@ pub fn register_event_handler(
     evhdlr: NotificationFn,
     cbfunc: HandlerRegCbFn,
 ) -> Result<EventHandlerRef, PmixStatus> {
+    // Serialize blocking registrations so PENDING_REGISTRATIONS has one slot.
     let _registration_guard = BLOCKING_REGISTRATION_LOCK
         .lock()
         .unwrap_or_else(|e| e.into_inner());
@@ -516,10 +533,11 @@ pub fn register_event_handler(
     if status.is_success() {
         let handler_ref = raw_status as EventHandlerRef;
         if evhdlr.is_some() {
-            // Insert before removing the provisional entry. The bridge checks
-            // HANDLER_REGISTRY before PENDING_REGISTRATIONS, so this ordering
-            // leaves no re-key gap. The registry and pending locks are never
-            // nested in the bridge, avoiding an ABBA cycle.
+            // Insert before removing the provisional entry. Together with the
+            // bridge's post-pending registry re-read, this ensures a completed
+            // re-key is observed even if the bridge's first registry read
+            // preceded this critical section. The registry and pending locks
+            // are never nested in the bridge, avoiding an ABBA cycle.
             let mut pending = PENDING_REGISTRATIONS
                 .lock()
                 .expect("mutex poisoned (events.rs)");
@@ -1118,6 +1136,8 @@ mod tests {
                 user_fn: Box::new(Some(handler)),
             });
         let chain = 0xF_u8;
+        // SAFETY: test invokes the bridge with null FFI pointers and a live
+        // stack chain token; the completion callback is valid for this call.
         unsafe {
             notification_bridge(
                 424252,
@@ -1147,6 +1167,15 @@ mod tests {
                 .expect("mutex poisoned (events.rs)")
                 .contains_key(&424252)
         );
+        // Leave a non-matching pending entry to force the post-pending
+        // registry re-read on this delivery; the re-keyed handler must run.
+        PENDING_REGISTRATIONS
+            .lock()
+            .expect("mutex poisoned (events.rs)")
+            .push(PendingRegistration {
+                codes: vec![18],
+                user_fn: Box::new(Some(handler)),
+            });
         // SAFETY: test invokes the bridge with null FFI pointers and a live
         // stack chain token; the completion callback is valid for this call.
         unsafe {
@@ -1208,6 +1237,8 @@ mod tests {
             .lock()
             .expect("mutex poisoned (events.rs)")
             .insert(ref_id);
+        // SAFETY: test invokes the bridge with null FFI pointers; the
+        // completion callback is valid for this call.
         unsafe {
             notification_bridge(
                 ref_id,
--- a/src/events.rs
+++ b/src/events.rs
@@ -53,7 +53,7 @@
 //! deregister_event_handler(handler_ref, None).expect("deregister failed");
 //! ```
 
-use crate::{ffi, Info, PmixDataRange, PmixError, PmixStatus, Proc};
+use crate::{Info, PmixDataRange, PmixError, PmixStatus, Proc, ffi};
 use std::collections::HashMap;
 use std::os::raw::c_void;
 use std::ptr;
@@ -195,6 +195,14 @@ type HandlerRegistry = HashMap<EventHandlerRef, Box<NotificationFn>>;
 static HANDLER_REGISTRY: LazyLock<Mutex<HandlerRegistry>> =
     LazyLock::new(|| Mutex::new(HashMap::new()));
 
+struct PendingRegistration {
+    codes: Vec<i32>,
+    user_fn: Box<NotificationFn>,
+}
+
+static PENDING_REGISTRATIONS: LazyLock<Mutex<Vec<PendingRegistration>>> =
+    LazyLock::new(|| Mutex::new(Vec::new()));
+
 /// Drop every parked notification handler.
 ///
 /// Called from client / tool / server session finalize paths so registrations
@@ -202,6 +210,10 @@ static HANDLER_REGISTRY: LazyLock<Mutex<HandlerRegistry>> =
 /// for the rest of the process lifetime. Safe under the crate's no-reinit
 /// policy: after finalize, no further event deliveries are expected.
 pub(crate) fn clear_handler_registry() {
+    PENDING_REGISTRATIONS
+        .lock()
+        .unwrap_or_else(|e| e.into_inner())
+        .clear();
     HANDLER_REGISTRY
         .lock()
         .unwrap_or_else(|e| e.into_inner())
@@ -251,11 +263,10 @@ unsafe fn complete_event_chain(
 /// waits on that completion, so a missed `cbfunc` is a permanent hang — not
 /// a silent drop.
 ///
-/// On registry miss (no user fn, `evhdlr = None`, deregistration race, or
-/// the brief window after blocking registration returns but before the
-/// insert) this bridge still calls `cbfunc` as a pass-through so the chain
-/// stays alive. The chain pointer is always forwarded verbatim — never
-/// replaced with `NULL`.
+/// On registry miss (no user fn, `evhdlr = None`, deregistration race, or an
+/// unknown refid), this bridge still calls `cbfunc` as a pass-through. During
+/// blocking registration, a matching provisional handler is used while its
+/// refid is being returned by PMIx.
 ///
 /// User-handler panics are caught: unwinding across the C→Rust FFI boundary
 /// is undefined behaviour, and a panic that skips `cbfunc` would stall the
@@ -280,7 +291,18 @@ unsafe extern "C" fn notification_bridge(
         registry
             .get(&evhdlr_registration_id)
             .and_then(|b| *b.as_ref())
-    };
+    }
+    .or_else(|| {
+        let pending = PENDING_REGISTRATIONS
+            .lock()
+            .unwrap_or_else(|e| e.into_inner());
+        pending
+            .iter()
+            .find(|registration| {
+                registration.codes.is_empty() || registration.codes.contains(&status)
+            })
+            .and_then(|registration| *registration.user_fn.as_ref())
+    });
 
     if let Some(user_fn) = user_fn {
         // SAFETY: cast FFI pointers to c_void for the user-facing NotificationFn
@@ -397,14 +419,9 @@ extern "C" fn handler_reg_cb_bridge(status: i32, refid: EventHandlerRef, cbdata:
 ///
 /// # Registration-window delivery
 ///
-/// The C-side handler is live as soon as `PMIx_Register_event_handler`
-/// returns, but the reference ID is only known **after** that return, so
-/// the Rust registry insert cannot run earlier. An event delivered in that
-/// gap is **pass-through-completed without invoking `evhdlr`** (dropped at
-/// the Rust layer; the OpenPMIx chain still advances). Loss-sensitive
-/// notifications (e.g. `PMIX_EVENT_JOB_END`) can therefore miss a delivery
-/// that races registration — re-check state after register returns, or
-/// tolerate a possible drop in that window.
+/// The handler is parked provisionally before entering PMIx, so a matching
+/// event delivered before the returned reference ID can be re-keyed is
+/// dispatched to `evhdlr` rather than dropped.
 ///
 /// # C API
 /// ```c
@@ -449,6 +466,16 @@ pub fn register_event_handler(
         (ptr::null(), 0)
     };
 
+    if evhdlr.is_some() {
+        PENDING_REGISTRATIONS
+            .lock()
+            .expect("mutex poisoned (events.rs)")
+            .push(PendingRegistration {
+                codes: codes.iter().map(|status| status.to_raw()).collect(),
+                user_fn: Box::new(evhdlr),
+            });
+    }
+
     // SAFETY: FFI call into PMIx library. The codes slice and info handle
     // remain valid for the duration of this call. The user's notification fn
     // is kept alive by HANDLER_REGISTRY (keyed by the reference ID PMIx
@@ -470,20 +497,25 @@ pub fn register_event_handler(
     if status.is_success() {
         let handler_ref = raw_status as EventHandlerRef;
         if evhdlr.is_some() {
-            // Park the user fn for the lifetime of the registration.
-            //
-            // Race window: the C-side handler is live as soon as
-            // PMIx_Register_event_handler returns, but the refid is only known
-            // *after* that return, so the insert cannot move earlier. An event
-            // delivered in this gap hits a registry miss in notification_bridge,
-            // which pass-through-completes the chain (dropped delivery, no hang).
+            let user_fn = PENDING_REGISTRATIONS
+                .lock()
+                .expect("mutex poisoned (events.rs)")
+                .pop()
+                .expect("pending blocking registration missing")
+                .user_fn;
             HANDLER_REGISTRY
                 .lock()
                 .expect("mutex poisoned (events.rs)")
-                .insert(handler_ref, Box::new(evhdlr));
+                .insert(handler_ref, user_fn);
         }
         Ok(handler_ref)
     } else {
+        if evhdlr.is_some() {
+            PENDING_REGISTRATIONS
+                .lock()
+                .expect("mutex poisoned (events.rs)")
+                .pop();
+        }
         Err(status)
     }
 }
@@ -738,11 +770,7 @@ pub fn notify_event(
     };
 
     let st = PmixStatus::from_raw(raw_status);
-    if st.is_success() {
-        Ok(())
-    } else {
-        Err(st)
-    }
+    if st.is_success() { Ok(()) } else { Err(st) }
 }
 
 /// Non-blocking variant of [`notify_event`].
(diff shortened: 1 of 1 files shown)


src/events.rs
```diff
@@ -470,20 +502,25 @@ pub fn register_event_handler(
...
         let handler_ref = raw_status as EventHandlerRef;
         if evhdlr.is_some() {
-            // Park the user fn for the lifetime of the registration.
-            //
-            // Race window: the C-side handler is live as soon as
-            // PMIx_Register_event_handler returns, but the refid is only known
-            // *after* that return, so the insert cannot move earlier. An event
-            // delivered in this gap hits a registry miss in notification_bridge,
-            // which pass-through-completes the chain (dropped delivery, no hang).
+            let user_fn = PENDING_REGISTRATIONS
```
SedahsReview: Warning — the re-key is not atomic with respect to the bridge's pending consult, so the drop window from #391 is narrowed but not closed. `PENDING_REGISTRATIONS.lock()...pop()` (guard dropped at the end of this statement) and `HANDLER_REGISTRY.lock()...insert()` are two separate critical sections. A delivery for the just-returned refid that lands between them sees neither the registry entry (not yet inserted) nor the pending entry (already popped) and is pass-through-completed — the first event is still silently dropped, which is exactly the class of loss this PR sets out to eliminate. The gap is sub-microsecond but real: for blocking registration OpenPMIx gives no "no delivery before the call returns" guarantee (that guarantee exists only for the non-blocking form), so the bridge can fire at any point after the C-side registration becomes live.

Fix (deadlock-free — notification_bridge never holds HANDLER_REGISTRY while acquiring PENDING, its registry guard is scoped to the first block, so there is no ABBA cycle): hold the PENDING guard across the insert, e.g.

    let mut pending = PENDING_REGISTRATIONS.lock().expect("mutex poisoned (events.rs)");
    let user_fn = pending.pop().expect("pending blocking registration missing").user_fn;
    HANDLER_REGISTRY.lock().expect("mutex poisoned (events.rs)").insert(handler_ref, user_fn);

or reorder to insert-then-pop, which leaves no gap at all (registry lookup precedes the pending consult, so the brief pending linger is harmless).

src/events.rs
```diff
@@ -280,7 +292,18 @@ unsafe extern "C" fn notification_bridge(
...
-    };
+    }
+    .or_else(|| {
+        let pending = PENDING_REGISTRATIONS
+            .lock()
+            .unwrap_or_else(|e| e.into_inner());
+        pending
+            .iter()
+            .find(|registration| {
+                registration.codes.is_empty() || registration.codes.contains(&status)
```
SedahsReview: Warning — the pending lookup matches by codes only and cannot distinguish the pending registration's own delivery from another registration's delivery that happens to land in the same window. Two concrete triggers:

1. Deregistration race: deregister_event_handler removes the refid from HANDLER_REGISTRY (line ~664) *before* PMIx_Deregister_event_handler returns, and the C-side handler stays live until that C call completes. A bridge invocation for the dying refid in that window hits a registry miss and falls into this pending consult; if the event's status matches the pending codes, the pending handler is invoked with the *wrong* refid, and when the pending registration's own bridge call also fires for the same event the handler is double-delivered.

2. Empty-codes entries: `codes.is_empty()` matches ANY status. PMIx semantics for NULL/0 codes is a *default* handler invoked only for events not claimed by a code-specific handler (docs.openpmix.org PMIx_Register_event_handler man page) — so the match-all is correct for the default handler's own deliveries, but it also lets a pending default registration hijack any dying-registration delivery regardless of codes.

No hang, no use-after-free, and no single chain is double-completed (each bridge invocation carries its own chain), so this is not memory-unsafe — but the user-visible contract (handler invoked once per matching event, with its own refid) can be violated in a narrow 3-way race (in-flight blocking registration + concurrent deregistration of a codes-overlapping handler + event). At minimum this limitation should be documented in the bridge doc (the "deregistration race → pass-through" statement is now only conditionally true when a pending registration with matching codes exists); ideally the pending consult should be constrained (e.g., only consult pending for deliveries whose refid cannot be a live registration) or the window closed by deferring the registry removal until after the C deregister completes.

SedahsReview (changes requested):
Reviewed locally (pr-393 @ 2f0d186, 2 commits: window-closing + serialization lock). The pending-registration mechanism is a sound, well-documented approach to issue #391 — but as written it does not fully close the drop window (the re-key is not atomic with respect to the bridge's pending consult), and the codes-only pending match can misdeliver another registration's delivery to the pending handler in a narrow race. Both issues are concrete and fixable; details inline.

Quality gates (PMIX_PREFIX=$HOME/.local/openpmix-6.1.0): cargo build OK; cargo test --lib events::tests -- --test-threads=1 -> 125 passed / 0 failed; cargo clippy --lib -- -D warnings -> clean; cargo fmt --check -> src/events.rs is fmt-clean (repo carries 589 pre-existing fmt diffs elsewhere, none from this PR).

SedahsReview:
## Review Summary (SedahsReview)

**Verdict: REQUEST_CHANGES** — 2 warnings (both inline on src/events.rs), plus suggestions. The pending-registration mechanism is a sound, well-documented approach to #391 and the push/pop pairing, lock discipline, and docs are all correct; the blocking issues are the residual (unclosed) drop window and the codes-only pending match.

### Critical
- None. No memory-safety, hang, or single-chain double-completion issues found.

### Warnings
- **src/events.rs:505** — The re-key is not atomic w.r.t. the bridge's pending consult: PENDING_REGISTRATIONS.pop() (lock released at statement end) and HANDLER_REGISTRY.insert() are two separate critical sections. A delivery for the just-returned refid landing between them sees neither entry and is pass-through-completed — the #391 drop window is narrowed to a sub-microsecond gap but not closed. Fix: hold the PENDING lock across the insert (deadlock-free, since the bridge never nests HANDLER->PENDING), or insert-then-pop.
- **src/events.rs:303** — The pending lookup matches by codes only and cannot tell the pending registration's own delivery apart from another registration's delivery: a bridge invocation for a dying refid during a deregistration race (registry removed before the C deregister returns) can be misdelivered to the pending handler with the wrong refid, and double-delivered when the pending registration's own bridge call also fires. codes.is_empty() (default handler, PMIx NULL/0-codes semantics) matches any status, widening this. Fix or document the limitation.

### Suggestions
- **Test coverage** — test_pending_registration_delivery_and_rekey exercises the bridge pending-consult path and manual re-key (deterministic, exactly-once user-fn invocation verified via CALLED==1), but register_event_handler's push/pop pairing is never exercised: mock_ffi has no PMIx_Register_event_handler, so the success/error/evhdlr=None/codes-empty paths of the new logic are untested. Consider a mock for the FFI call or a refactor that isolates the pairing logic for unit testing.
- **Scope creep (nit)** — the rustfmt collapse of the notify_event/notify_event_nb bodies and two pre-existing tests is unrelated churn; behavior is identical, but it belongs in a separate commit.
- **Robustness (nit)** — clear_handler_registry clearing PENDING while a blocking registration is in flight (concurrent finalize + register, out-of-contract misuse) would make the post-FFI .expect("pending blocking registration missing") panic; a panic-free pop would be more defensive.

### Looks Good
- NotificationFn is Option<fn-ptr> (Copy), so the evhdlr.is_some() checks after Box::new(evhdlr) are sound and push/pop are correctly paired on every in-contract path (success, error, evhdlr=None).
- Lock discipline is correct: BLOCKING_REGISTRATION_LOCK serializes blocking registrations (at-most-one pending entry invariant holds); no lock is held across the user callback; no mutex nesting anywhere (bridge: HANDLER-scoped then PENDING-scoped; register: PENDING then HANDLER, guards non-nested) — no ABBA deadlock.
- Deregistration ordering (registry remove before C call) preserved; pending entries can't resurrect freed handlers; finalize clears pending Boxes.
- Bridge and registration-window docs updated accurately; no stale "drop window" text; no secrets/conflict markers/debug leftovers; all new unsafe blocks follow the file's existing SAFETY-comment convention.
- Test uses MockGuard consistently and is deterministic (single-threaded, no flakes).

### Gates (local, PMIX_PREFIX=$HOME/.local/openpmix-6.1.0)
- cargo build — OK
- cargo test --lib events::tests -- --test-threads=1 — 125 passed, 0 failed
- cargo clippy --lib -- -D warnings — clean
- cargo fmt --check — src/events.rs fmt-clean; 589 pre-existing repo-wide diffs (examples etc.), none attributable to this PR

src/events.rs
```diff
@@ -470,20 +516,32 @@ pub fn register_event_handler(
...
-            // Race window: the C-side handler is live as soon as
-            // PMIx_Register_event_handler returns, but the refid is only known
-            // *after* that return, so the insert cannot move earlier. An event
-            // delivered in this gap hits a registry miss in notification_bridge,
-            // which pass-through-completes the chain (dropped delivery, no hang).
+            // Insert before removing the provisional entry. The bridge checks
+            // HANDLER_REGISTRY before PENDING_REGISTRATIONS, so this ordering
+            // leaves no re-key gap. The registry and pending locks are never
+            // nested in the bridge, avoiding an ABBA cycle.
+            let mut pending = PENDING_REGISTRATIONS
```
SedahsReview, line 541: Residual drop window: the insert-then-pop under the PENDING guard is a big improvement, but the window is not fully closed. The bridge reads HANDLER_REGISTRY first (lines 297-302), then — after a miss — PENDING (lines 311-319). Those two reads can straddle this whole critical section:

1. Register thread: locks PENDING (line 523), then blocks acquiring HANDLER_REGISTRY while the bridge (progress thread) currently holds it.
2. Bridge: reads HANDLER_REGISTRY → miss (insert not yet done), releases.
3. Register: acquires HANDLER_REGISTRY, inserts, pops, releases PENDING.
4. Bridge: checks DEREG_IN_PROGRESS (miss), then locks PENDING → reads an empty vec (already popped) → pass-through → **dropped delivery**.

No exotic timing is required: the register grabs PENDING immediately on C-return and can be blocked on the registry lock the bridge holds, and the bridge then blocks on PENDING until the re-key completes. This is exactly the first-event race the PR claims to close.

Provably complete one-line fix: in the bridge's `or_else`, when the pending consult finds nothing, re-read HANDLER_REGISTRY — pending empty ⇒ pop already done ⇒ insert already done (same critical section) ⇒ the re-read must hit. Cost: one extra registry lock acquisition on the miss path only. Alternatively, hold the PENDING guard across a registry re-read.

Also: the comment at lines 519-522 ('this ordering leaves no re-key gap') is not strictly correct and should be updated to match whatever guarantee the code actually provides.

SedahsReview (changes requested):
Re-review of the fix for #391. Verdict: **REQUEST_CHANGES** — one residual race remains in the re-key path (warning 1 is not fully closed; the code's own comment overclaims), everything else checks out.

**Warning 2 (DEREG_IN_PROGRESS) — RESOLVED, verified.** The bridge consults DEREG_IN_PROGRESS before PENDING with each lock scoped and released (lines 303-320), so no nesting and no deadlock against the dereg path's DEREG_IN_PROGRESS→HANDLER_REGISTRY order. Both dereg paths insert the refid before the C call and remove it after `PmixStatus::from_raw` and before the `is_success` check, so the remove runs on both success and error branches. HANDLER_REGISTRY removal still precedes the C call, so the Box is dropped before the C-side handler goes dead; a delivery in that window hits registry-miss → DEREG_IN_PROGRESS-hit → pass-through and never touches the freed Box. `clear_handler_registry()` clears DEREG_IN_PROGRESS, PENDING_REGISTRATIONS and HANDLER_REGISTRY. Bridge docs (lines 273-277) match the code. The empty-codes default-handler hijack is now impossible for deregistering refids; the residual (a default pending registration claiming a delivery for a refid neither registered nor in DEREG_IN_PROGRESS) exists only while a blocking registration is in flight, delivers an event the default handler is entitled to, and is strictly better than dropping — see inline suggestion.

**Warning 1 (re-key atomicity) — substantially improved but NOT fully closed.** The insert-then-pop under the PENDING guard closes the dominant window, and the ABBA analysis is clean (all nested acquisitions take HANDLER_REGISTRY last; the bridge never holds two locks). The `.expect` calls are unreachable (push only occurs when `evhdlr.is_some()`), and the error-path `pop()` is correct under BLOCKING_REGISTRATION_LOCK. However, a delivery can still be dropped by the lock-interleaving detailed in the inline comment: the bridge's two reads (HANDLER_REGISTRY, then PENDING) can straddle the entire re-key critical section because the re-key holds PENDING (needed by the bridge's second read) while the bridge's first read already missed. The comment at lines 519-522 ('leaves no re-key gap') is therefore not strictly correct.

**Quality gates:** `cargo build` ✓ · `cargo test --lib events::tests -- --test-threads=1` ✓ (126 passed, incl. both new tests) · `cargo clippy --lib -- -D warnings` ✓ · `cargo fmt --check` — src/events.rs clean (589 pre-existing diffs elsewhere, not PR-attributable) · `cargo check/test --features mock_ffi` ✓ (44 baseline warnings).

SedahsReview:
## Code Review Summary — Re-review of #393 (fix for #391)

**Verdict: Changes Requested** (1 warning remains; 1 warning resolved; 3 suggestions)

### Warnings
- **src/events.rs:523 (re-key)** — Residual drop window remains despite insert-then-pop. The bridge's two reads (HANDLER_REGISTRY at 297-302, then PENDING at 311-319) can straddle the whole re-key critical section: register locks PENDING, blocks on HANDLER_REGISTRY held by the bridge; bridge reads registry → miss; register inserts, pops, releases PENDING; bridge reads pending → empty → pass-through → delivery dropped. This is the exact first-event race the PR claims to close. Fix (one line, provably complete): re-read HANDLER_REGISTRY after a pending miss (pending empty ⇒ pop done ⇒ insert done ⇒ re-read hits). The comment at 519-522 ('leaves no re-key gap') is not strictly correct.

### ✅ Resolved (previous warning 2)
- **DEREG_IN_PROGRESS gate** — Verified correct: bridge consults it before PENDING, locks never nested (no deadlock with dereg's DEREG→HANDLER order); both dereg paths insert before the C call and remove after `PmixStatus::from_raw` on success **and** error paths; HANDLER_REGISTRY removal still precedes the C call (no UAF — the bridge copies a plain fn pointer out under the lock); `clear_handler_registry()` clears all three structures; bridge docs match the code.

### 💡 Suggestions
- **src/events.rs:1121, 1211** — Two new `unsafe` blocks in the tests lack `// SAFETY:` comments (line 1150 has one — inconsistent). Add them for uniformity.
- **src/events.rs:462** — `_registration_guard` (BLOCKING_REGISTRATION_LOCK) serializes all blocking registrations; a one-line comment explaining why (single pending slot, `pending.last()` assumes exclusivity) would help future readers.
- **src/events.rs:273-277** — Provisional-match doc could note the foreign-refid caveat: while a blocking registration is in flight, a default (empty-codes) pending registration can claim a delivery for a refid that is neither registered nor in DEREG_IN_PROGRESS (the event is one it registered for; the refid argument may be foreign). Narrow, inherent to the design, strictly better than dropping.

### ✅ Looks Good
- Push/pop pairing on all paths (push before C call; insert-then-pop on success; pop on error; all gated by `evhdlr.is_some()`).
- `.expect("pending blocking notification fn missing")` unreachable in production (push only when `evhdlr.is_some()`; `NotificationFn` is Copy).
- No ABBA cycle: all nested acquisitions take HANDLER_REGISTRY last; the bridge never holds two locks.
- Chain completed exactly once; no lock held across user callbacks; panic containment via `catch_unwind` intact.
- Both new tests deterministic (statics + MockGuard, cleared registry between phases, `--test-threads=1`).
- No secrets, debug leftovers, or conflict markers in the diff.

### Quality gates
- `cargo build` — PASS
- `cargo test --lib events::tests -- --test-threads=1` — PASS (126 passed, incl. `test_pending_registration_delivery_and_rekey` and `test_pending_consult_skipped_for_deregistering_ref`)
- `cargo clippy --lib -- -D warnings` — PASS (clean)
- `cargo fmt --check` — src/events.rs clean (589 pre-existing diffs elsewhere, none PR-attributable)
- `cargo check/test --features mock_ffi --lib` — PASS (44 warnings, baseline)

src/events.rs
```diff
@@ -1000,6 +1087,175 @@ mod tests {
...
+            .lock()
+            .expect("mutex poisoned (events.rs)")
+            .insert(424252, pending.user_fn);
+        assert!(
+            HANDLER_REGISTRY
+                .lock()
+                .expect("mutex poisoned (events.rs)")
+                .contains_key(&424252)
+        );
+        // Leave a non-matching pending entry to force the post-pending
```
SedahsReview: Warning (non-blocking, test fidelity): this comment overclaims — the second bridge call does not exercise the post-pending registry re-read. The test inserts refid 424252 into HANDLER_REGISTRY at lines 1160-1163, so the bridge's FIRST registry read (lines 299-304) hits and the or_else chain (DEREG check → pending consult → re-read) never runs; the non-matching pending entry (codes [18]) is never consulted. The re-read's HIT branch — the exact guarantee that closes #391 — therefore has no test coverage: this test passes even if lines 324-335 are deleted. (The re-read's miss branch is covered by test_notification_bridge_unknown_ref_invokes_cbfunc.) The straddle closure is an ordering property (insert before pop under one PENDING critical section), so it cannot be exercised deterministically single-threaded. Please either correct the comment to state what the test actually verifies (re-keyed delivery resolves via the registry path), or add a test seam / threaded test that lands the registry insert between the bridge's two reads. Production code is correct as-is; this does not block merge.

SedahsReview (approved):
Re-review (round 3) of the fix for #391. Verdict: **APPROVE** — the residual re-key straddle from round 2 is genuinely closed by the post-pending registry re-read; no new critical or warning-level production issue. One non-blocking test-fidelity warning is noted inline (the new test does not actually exercise the re-read hit branch and its comment overclaims).

**Round-2 blocking warning (re-key straddle) — RESOLVED, verified.** The bridge now does: HANDLER_REGISTRY read (scoped) → DEREG_IN_PROGRESS check (scoped) → PENDING consult (scoped) → on pending miss, a second HANDLER_REGISTRY read (scoped). Walking every interleaving: the re-key holds PENDING_REGISTRATIONS for the entire insert→pop, so the bridge's pending consult either (a) runs before the re-key and finds the entry (codes match → delivered; codes mismatch → the delivery is not for the in-flight registration, pass-through correct), or (b) runs after the re-key released PENDING, in which case the pop is done and the insert — which precedes the pop in program order and released the registry lock before PENDING — is visible to the re-read. The bridge can never observe "pending empty but insert not yet done", because the re-key holds PENDING until after the insert. A re-read hit is always a legitimate delivery (PMIx invokes the bridge only for a handler whose registered codes match the event, so a by-refid hit is authoritative). No ordering drops a first-event delivery belonging to the just-registered handler. The overclaiming comment is corrected (lines 536-540).

**Lock ordering — still no-ABBA.** All nested acquisitions take HANDLER_REGISTRY last (register re-key: PENDING→HANDLER_REGISTRY; dereg: DEREG_IN_PROGRESS→HANDLER_REGISTRY); the bridge holds at most one lock at a time (four scoped reads, never nested); clear_handler_registry is sequential. No cycle.

**Re-read is strictly on the miss+miss path** — the hot path (registry hit) is unchanged; one extra registry lock only after a registry miss + dereg miss + pending miss.

**Round-2 suggestions — applied:** `// SAFETY:` comments on the test `unsafe` blocks (lines 1139-1140, 1179-1180, 1240-1241); `_registration_guard` serialization comment (line 478); provisional-match foreign-refid doc caveat (lines 278-279).

**Previously approved items — still hold:** push/pop pairing on success and error paths; DEREG_IN_PROGRESS removed on success and error in both dereg variants; registry-remove-before-C-call UAF ordering; clear_handler_registry clears all three structures; chain completed exactly once per delivery; no lock held across user callbacks; panic containment with chain recovery; `// SAFETY:` on all `unsafe`; deterministic single-threaded tests; no secrets/debug leftovers/conflict markers.

**New finding (non-blocking, test fidelity):** in `test_pending_registration_delivery_and_rekey`, the second bridge call does not exercise the post-pending registry re-read: the test inserts refid 424252 into HANDLER_REGISTRY before the call (lines 1160-1163), so the bridge's first registry read hits and the or_else chain never runs. The comment at lines 1170-1171 ("force the post-pending registry re-read") is inaccurate. The re-read's miss branch is covered by `test_notification_bridge_unknown_ref_invokes_cbfunc`; the hit branch (the actual #391 guarantee) is not — the test passes even if the re-read is deleted. Since the straddle closure is an ordering property (insert before pop under one PENDING critical section), a deterministic single-threaded test of the hit branch requires a test seam or a threaded race test; at minimum the comment should describe what the test actually verifies (re-keyed delivery resolves via the registry path).

**Quality gates:** cargo build ✓ · cargo test --lib events::tests -- --test-threads=1 ✓ (126 passed) · cargo clippy --lib -- -D warnings ✓ · cargo fmt --check: src/events.rs clean (589 pre-existing diffs elsewhere, not PR-attributable) · cargo check --features mock_ffi --lib ✓ (44 baseline warnings).

SedahsReview:
**Re-review (round 3) — Verdict: APPROVE** ✅ (fix for #391 is complete; one non-blocking test-fidelity warning)

**Critical** — none.

**Warnings**
- src/events.rs:1170-1193 (test) — `test_pending_registration_delivery_and_rekey` does not exercise the post-pending registry re-read: the test inserts refid 424252 into HANDLER_REGISTRY (lines 1160-1163) before the second bridge call, so the first registry read hits and the re-read never runs. The comment "force the post-pending registry re-read" is inaccurate, and the re-read HIT branch (the #391 guarantee) has no regression test — this test passes even if the re-read is deleted. The re-read miss branch is covered by `test_notification_bridge_unknown_ref_invokes_cbfunc`. Suggest correcting the comment, or adding a seam/threaded test that lands the registry insert between the bridge's two reads. Non-blocking.

**Suggestions** — none new.

**Round-2 items — all resolved**
- Re-key straddle: closed by the post-pending registry re-read; verified by exhaustive interleaving (see the formal review).
- Comment overclaim (old lines 519-522): corrected (lines 536-540).
- `// SAFETY:` comments on the new test unsafe blocks: present (1139-1140, 1179-1180, 1240-1241).
- `_registration_guard` serialization comment: present (line 478).
- Provisional-match foreign-refid doc caveat: present (lines 278-279).

**Looks Good**
- Lock ordering: no ABBA possible — all nested acquisitions take HANDLER_REGISTRY last; the bridge never holds two locks.
- Re-read fires only on the registry-miss + pending-miss path; the hot path is unchanged.
- push/pop pairing on success and error; DEREG_IN_PROGRESS cleared on success and error; registry-remove-before-C-call UAF ordering; clear_handler_registry clears all three structures; chain completed exactly once; no lock across user callbacks; panic containment with chain recovery; `// SAFETY:` on all unsafe; deterministic single-threaded tests; no secrets/debug leftovers/conflict markers.

**Quality gates**
- cargo build ✓
- cargo test --lib events::tests -- --test-threads=1 ✓ (126 passed)
- cargo clippy --lib -- -D warnings ✓
- cargo fmt --check — src/events.rs clean (589 pre-existing diffs elsewhere, not PR-attributable)
- cargo check --features mock_ffi --lib ✓ (44 baseline warnings)

Reviewed by Hermes Agent (SedahsReview).

PR #395, opened 2026-08-08
fix(data): reclaim pmix_value_t struct allocation in get_value/get_nb

### LINKED ISSUE #389: bug: get_value/get_nb leak the pmix_value_t struct allocation; get_nb bridge writes zeroed() into C-owned memory
## Summary

The value-ownership handoff for `get_value` (blocking) and `get_nb` does not match OpenPMIx's release contract, causing a per-call leak of the `pmix_value_t` struct allocation — and the `get_nb` bridge additionally writes into C-owned memory, which is UB-adjacent.

## Evidence (verified against openpmix v6.1.0 source)

**Blocking `PMIx_Get`** hands over a heap-allocated struct: `src/client/pmix_client_get.c` — `*val = cb->value; cb->value = NULL;` where `cb->value` was allocated via `PMIX_VALUE_CREATE(cb->value, 1)` (`:884`, `:987`, `:1054`). Standard contract: the caller must release with `PMIX_VALUE_RELEASE(*val)`.

The crate does `PmixOwnedValue { inner: unsafe { *value } }` (`src/lib.rs:3868-3872`) — a **struct copy**. `PmixOwnedValue::Drop` → `free_value(&mut self.inner)` (`src/value.rs:703-707`) frees **only nested payloads** (CString/Box/Vec arms, `:576-635`), never the struct itself. The heap struct from C is therefore leaked on every successful blocking `get_value`.

**`PMIx_Get_nb`**: the completion hands `kv` to the callback; the `pmix_cb_t` destructor (`cbdes` in `src/common/pmix_globals.c`) does **not** release `p->value`, so the struct allocation is also never freed by the library. The bridge (`src/data_ops/mod.rs:258-277`) does:

```rust
let val = unsafe { ptr::read(kv) };            // struct copy
unsafe { ptr::write(kv, std::mem::zeroed()) }; // ← writes into C-owned memory
```

- The zeroing is a defensive hack ("so PMIx doesn't double-free") that depends on OpenPMIx never reading `*kv` again after the callback returns — true today, but an undocumented, version-sensitive invariant.
- In the `PMIX_QUALIFIED_VALUE` path (`pmix_client_get.c:714-731`), `cb->value` points **into** a C-owned `pmix_info_t` array (`cb->value = &iptr[0].value`). Zeroing that slot mutates C-owned memory, and `free_value` on the copy will free nested pointers that may alias the C allocation → heap corruption / double-free risk.
- The struct allocation itself still leaks on the plain path.

The crate's own doc comment admits the hazard: `src/value.rs:573-575` — "Calling this on a pmix_value_t produced by the C library (and therefore managed by PMIX_VALUE_RELEASE) is a double-free."

## Suggested fix

- After `ptr::read(kv)` (or `*value`), reclaim the C struct allocation: `drop(Box::from_raw(value))` / call `PMIX_VALUE_RELEASE` on the original pointer so struct + payload are freed exactly once by the C convention, and keep the Rust copy for the RAII value.
- Never `ptr::write` into memory owned by the C library; treat `kv` as read-only.
- Add a valgrind/ASan run of `tests/data_ops_Get_nb.rs` + the daemon get tests to prove zero leaks.

This is observable in long-running HPC clients doing many gets: one ~40–48 byte struct leaked per successful get.

SedahsDev:
Closes #389

## Summary
- Reclaims the heap-allocated pmix_value_t struct returned by blocking PMIx_Get after copying its payload pointers into the Rust RAII value.
- Reclaims the plain PMIx_Get_nb struct allocation without writing zeroed data into C-owned memory.
- Detects PMIX_QUALIFIED_VALUE directives and deep-copies through PMIx_Value_xfer, leaving the C-owned info array untouched.

The installed OpenPMIx 6.1.0 headers confirm PMIX_QUALIFIED_VALUE is the `pmix.qual.val` directive. The installed headers expose the pmix_cb_t type and callback API, but the distribution does not include OpenPMIx client source/cbdes implementation; the destructor claim was therefore not independently verified from installed source.

## Test plan
- `PMIX_PREFIX=$HOME/.local/openpmix-6.1.0 cargo build`
- `PMIX_PREFIX=$HOME/.local/openpmix-6.1.0 cargo test --lib data_ops::tests::test_get_value_callback_bridge_invokes_callback`
- Full daemon-free mock tests are constrained by pre-existing test compilation issues in the repository.
- `cargo clippy --all-targets -- -D warnings` remains blocked by pre-existing warnings/errors in mock/test sources.
- `cargo fmt --check` reports pre-existing formatting differences.

### DIFF
--- a/src/data_ops/mod.rs
+++ b/src/data_ops/mod.rs
@@ -7,7 +7,7 @@
 // PMIx_Publish — publish data for later lookup
 // ─────────────────────────────────────────────────────────────────────────────
 
-use std::ffi::CString;
+use std::ffi::{CStr, CString};
 use std::os::raw::c_void;
 use std::ptr;
 use std::sync::{LazyLock, Mutex};
@@ -219,6 +219,8 @@ pub trait GetValueCallback: Send {
 type GetRegistry = std::collections::HashMap<usize, Box<dyn GetValueCallback>>;
 static GET_REGISTRY: LazyLock<Mutex<GetRegistry>> =
     LazyLock::new(|| Mutex::new(std::collections::HashMap::new()));
+static QUALIFIED_GETS: LazyLock<Mutex<std::collections::HashSet<usize>>> =
+    LazyLock::new(|| Mutex::new(std::collections::HashSet::new()));
 
 /// Monotonically increasing request ID counter.
 static GET_SEQ: LazyLock<Mutex<usize>> = LazyLock::new(|| Mutex::new(0));
@@ -252,20 +254,37 @@ extern "C" fn get_value_callback_bridge(
         None => return, // Callback already consumed or never registered.
     };
 
+    let qualified = QUALIFIED_GETS
+        .lock()
+        .expect("mutex poisoned (data_ops.rs)")
+        .remove(&req_id);
+
     // Convert the status.
     let pmix_status = PmixStatus::from_raw(status);
 
-    // Extract the value on success.
+    // Extract the value on success. Qualified values borrow PMIx info memory.
     let value = if pmix_status.is_success() && !kv.is_null() {
-        // SAFETY: On success, PMIx returns a valid pmix_value_t that we
-        // take ownership of. We read it and then null the pointer so
-        // PMIx doesn't try to free it.
-        let val = unsafe { ptr::read(kv) };
-        // Clear the pointer so PMIx doesn't double-free.
-        unsafe { ptr::write(kv, std::mem::zeroed()) };
-        Some(PmixOwnedValue { inner: val, 
-            _not_thread_safe: std::marker::PhantomData,
-        })
+        if qualified {
+            // SAFETY: `kv` is valid during the callback; Value_xfer does not modify it.
+            let mut copied = unsafe { std::mem::zeroed() };
+            let copy_status = crate::pmix_ffi_or_mock!(
+                mock = unsafe { mock_ffi::mock_value_xfer(&mut copied, kv) },
+                real = unsafe { ffi::PMIx_Value_xfer(&mut copied, kv) },
+            );
+            (copy_status == ffi::PMIX_SUCCESS as i32).then_some(PmixOwnedValue {
+                inner: copied,
+                _not_thread_safe: std::marker::PhantomData,
+            })
+        } else {
+            // SAFETY: The ordinary callback path returns an allocated struct.
+            let val = unsafe { ptr::read(kv) };
+            // SAFETY: `kv` is the allocation returned by PMIx_Get_nb.
+            drop(unsafe { Box::from_raw(kv) });
+            Some(PmixOwnedValue {
+                inner: val,
+                _not_thread_safe: std::marker::PhantomData,
+            })
+        }
     } else {
         None
     };
@@ -342,6 +361,21 @@ pub fn get_nb(
         }
         None => (ptr::null(), 0),
     };
+
+    if let Some(info) = info {
+        if !info.handle.is_null() && info.len > 0 {
+            // SAFETY: `handle` points to the borrowed Info array.
+            let qualified = unsafe {
+                CStr::from_ptr((*info.handle).key.as_ptr()).to_bytes() == b"pmix.qual.val"
+            };
+            if qualified {
+                QUALIFIED_GETS
+                    .lock()
+                    .expect("mutex poisoned (data_ops.rs)")
+                    .insert(req_id);
+            }
+        }
+    }
 
     // Call the FFI function.
     let status = crate::pmix_ffi_or_mock!(
--- a/src/data_ops/mod.rs
+++ b/src/data_ops/mod.rs
@@ -236,6 +236,8 @@ extern "C" fn get_value_callback_bridge(
     cbdata: *mut c_void,
 ) {
     if cbdata.is_null() {
+        // Unreachable in practice: encode_req_id always supplies non-null cbdata,
+        // and PMIx invokes the callback with that same opaque pointer.
         return;
     }
 
--- a/src/data_ops/mod.rs
+++ b/src/data_ops/mod.rs
@@ -416,6 +416,12 @@ pub fn get_nb(
     } else {
         // Immediate failure — remove the registered callback so it
         // will never be invoked.
+        // The completion callback never runs on this path, so drop both the
+        // callback registration and any qualified-value marker.
+        QUALIFIED_GETS
+            .lock()
+            .expect("mutex poisoned (data_ops.rs)")
+            .remove(&req_id);
         let mut registry = GET_REGISTRY.lock().expect("mutex poisoned (data_ops.rs)");
         registry.remove(&req_id);
         Err(pmix_status)
--- a/src/data_ops/mod.rs
+++ b/src/data_ops/mod.rs
@@ -265,6 +265,10 @@ extern "C" fn get_value_callback_bridge(
     // Extract the value on success. Qualified values borrow PMIx info memory.
     let value = if pmix_status.is_success() && !kv.is_null() {
         if qualified {
+            // The server round-trip route provides a fresh heap struct that
+            // OpenPMIx does not reclaim, so this xfer leaves that struct and
+            // its nested payloads leaked. We cannot reclaim it here: the
+            // local-GDS route provides a pointer into C-owned info memory.
             // SAFETY: `kv` is valid during the callback; Value_xfer does not modify it.
             let mut copied = unsafe { std::mem::zeroed() };
             let copy_status = crate::pmix_ffi_or_mock!(
@@ -364,10 +368,13 @@ pub fn get_nb(
 
     if let Some(info) = info {
         if !info.handle.is_null() && info.len > 0 {
-            // SAFETY: `handle` points to the borrowed Info array.
-            let qualified = unsafe {
-                CStr::from_ptr((*info.handle).key.as_ptr()).to_bytes() == b"pmix.qual.val"
-            };
+            // SAFETY: `handle` points to `len` entries in the borrowed Info
+            // array, which remains valid for the duration of this call.
+            let entries = unsafe { std::slice::from_raw_parts(info.handle, info.len) };
+            let qualified = entries.iter().any(|entry| {
+                // SAFETY: each entry's key is a NUL-terminated pmix_key_t.
+                unsafe { CStr::from_ptr(entry.key.as_ptr()).to_bytes() == b"pmix.qual.val" }
+            });
             if qualified {
                 QUALIFIED_GETS
                     .lock()
--- a/src/data_ops/tests.rs
+++ b/src/data_ops/tests.rs
@@ -2060,15 +2060,19 @@ use super::*;
             registry.insert(req_id, Box::new(TestGetCb));
         }
 
-        // Create a mock pmix_value_t for the callback
-        let mut mock_value: ffi::pmix_value_t = unsafe { std::mem::zeroed() };
+        // Create a heap-allocated mock pmix_value_t for the callback.
+        // SAFETY: pmix_value_t is a C representation and zero is a valid
+        // initialization state before setting the type field.
+        let mut mock_value = Box::new(unsafe { std::mem::zeroed::<ffi::pmix_value_t>() });
         mock_value.type_ = PMIX_STRING_U16;
 
         let cbdata = crate::cbdata::encode_req_id(req_id);
         unsafe {
+            // SAFETY: The bridge takes ownership of the heap-allocated value,
+            // mirroring the real PMIx_Get_nb callback contract.
             get_value_callback_bridge(
                 PMIX_SUCCESS,
-                &mut mock_value as *mut ffi::pmix_value_t,
+                Box::into_raw(mock_value),
                 cbdata,
             );
         }
@@ -2795,15 +2799,19 @@ use super::*;
             registry.insert(req_id, Box::new(StringValueCb));
         }
 
-        // Create a mock pmix_value_t with PMIX_STRING type
-        let mut mock_value: ffi::pmix_value_t = unsafe { std::mem::zeroed() };
+        // Create a heap-allocated mock pmix_value_t with PMIX_STRING type.
+        // SAFETY: pmix_value_t is a C representation and zero is a valid
+        // initialization state before setting the type field.
+        let mut mock_value = Box::new(unsafe { std::mem::zeroed::<ffi::pmix_value_t>() });
         mock_value.type_ = PMIX_STRING_U16;
 
         let cbdata = crate::cbdata::encode_req_id(req_id);
         unsafe {
+            // SAFETY: The bridge takes ownership of the heap-allocated value,
+            // mirroring the real PMIx_Get_nb callback contract.
             get_value_callback_bridge(
                 PMIX_SUCCESS,
-                &mut mock_value as *mut ffi::pmix_value_t,
+                Box::into_raw(mock_value),
                 cbdata,
             );
         }
--- a/src/data_ops/tests.rs
+++ b/src/data_ops/tests.rs
@@ -2119,6 +2119,68 @@ use super::*;
         get_value_callback_bridge(PMIX_SUCCESS, std::ptr::null_mut(), std::ptr::null_mut());
     }
 
+    /// An immediate PMIx_Get_nb failure must reclaim the qualified marker.
+    #[test]
+    fn test_get_nb_failure_removes_qualified_marker() {
+        struct DummyGet;
+        impl GetValueCallback for DummyGet {
+            fn on_result(self: Box<Self>, _status: PmixStatus, _value: Option<PmixOwnedValue>) {}
+        }
+
+        let config = MockConfig::new().with_function_status("PMIx_Get_nb", PMIX_ERR_INIT);
+        let _guard = MockGuard::with_config(config);
+        let proc = Proc::new("mock.ns", 0).unwrap();
+        let mut builder = InfoBuilder::new();
+        builder.add_string_key("pmix.qual.val", "true", PMIX_STRING as _);
+        let info = builder.build();
+
+        let result = get_nb(&proc, "qualified.key", Some(&info), Box::new(DummyGet));
+
+        assert_eq!(result, Err(PmixStatus::Known(PmixError::ErrInit)));
+        assert!(QUALIFIED_GETS.lock().unwrap().is_empty());
+        assert!(GET_REGISTRY.lock().unwrap().is_empty());
+    }
+
+    /// A successful qualified callback uses mock value transfer and clears its marker.
+    #[test]
+    fn test_get_nb_qualified_success_uses_value_xfer() {
+        use std::sync::atomic::{AtomicBool, AtomicI32, Ordering};
+        static CB_STATUS: AtomicI32 = AtomicI32::new(-999);
+        static CB_HAS_VALUE: AtomicBool = AtomicBool::new(false);
+
+        struct QualifiedGet;
+        impl GetValueCallback for QualifiedGet {
+            fn on_result(self: Box<Self>, status: PmixStatus, value: Option<PmixOwnedValue>) {
+                CB_STATUS.store(status.to_raw(), Ordering::SeqCst);
+                CB_HAS_VALUE.store(value.is_some(), Ordering::SeqCst);
+            }
+        }
+
+        let _guard = MockGuard::new();
+        let req_id = {
+            let mut seq = GET_SEQ.lock().unwrap();
+            *seq += 1;
+            *seq
+        };
+        GET_REGISTRY
+            .lock()
+            .unwrap()
+            .insert(req_id, Box::new(QualifiedGet));
+        QUALIFIED_GETS.lock().unwrap().insert(req_id);
+
+        let mut mock_value = Box::new(unsafe { std::mem::zeroed::<ffi::pmix_value_t>() });
+        mock_value.type_ = PMIX_STRING_U16;
+        get_value_callback_bridge(
+            PMIX_SUCCESS,
+            Box::into_raw(mock_value),
+            crate::cbdata::encode_req_id(req_id),
+        );
+
+        assert_eq!(CB_STATUS.load(Ordering::SeqCst), PMIX_SUCCESS);
+        assert!(CB_HAS_VALUE.load(Ordering::SeqCst));
+        assert!(!QUALIFIED_GETS.lock().unwrap().contains(&req_id));
+    }
+
     // ─── Mock-aware lookup_nb callback tests ────────────────────────────────
 
     /// Test lookup_nb callback bridge with success status.
--- a/src/value.rs
+++ b/src/value.rs
@@ -570,9 +570,9 @@ fn write_payload(dst: &mut pmix_value, payload: PmixPayload) {
 /// After this call `v` is zeroed and safe to drop or reuse.
 ///
 /// # Safety
-/// `v` must have been produced by this crate's builder (or have equivalent
-/// allocation discipline).  Calling this on a `pmix_value_t` produced by the C
-/// library (and therefore managed by `PMIX_VALUE_RELEASE`) is a double-free.
+/// `v` must have been produced by this crate's builder, or be a Rust-owned copy
+/// whose nested payloads follow the same allocation discipline. It must not be
+/// a value still owned by PMIx or a borrowed copy of a C-owned info array.
 pub fn free_value(v: &mut pmix_value_t) {
     // SAFETY: type_ was set by write_payload; we access only the matching arm.
     unsafe {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2894,9 +2894,9 @@ fn write_payload(dst: &mut pmix_value, payload: PmixPayload) {
 /// After this call `v` is zeroed and safe to drop or reuse.
 ///
 /// # Safety
-/// `v` must have been produced by this crate's builder (or have equivalent
-/// allocation discipline).  Calling this on a `pmix_value_t` produced by the C
-/// library (and therefore managed by `PMIX_VALUE_RELEASE`) is a double-free.
+/// `v` must have been produced by this crate's builder, or be a Rust-owned copy
+/// whose nested payloads follow the same allocation discipline. It must not be
+/// a value still owned by PMIx or a borrowed copy of a C-owned info array.
 pub fn free_value(v: &mut pmix_value_t) {
     // SAFETY: type_ was set by write_payload; we access only the matching arm.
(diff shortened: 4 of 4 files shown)


src/data_ops/mod.rs
```diff
@@ -343,6 +362,21 @@ pub fn get_nb(
         None => (ptr::null(), 0),
     };
 
+    if let Some(info) = info {
+        if !info.handle.is_null() && info.len > 0 {
```
SedahsReview, line 372: **Blocking:** qualified detection only inspects `info[0]`. OpenPMIx treats info directives position-independently: the whole array is forwarded to the server (process_request scans *all* entries for the directives it handles client-side, and `PMIX_QUALIFIED_VALUE` is passed through unprocessed), so nothing requires the directive at index 0. If a caller builds e.g. `InfoBuilder::new().external_progress(true).add(b"pmix.qual.val", ...)`, the directive lands at index 1, PMIx honors it (returns a qualified value), but this bridge takes the **plain path** and does `Box::from_raw(kv)` on a pointer that points INTO the C-owned info array (`cb->value = &iptr[0].value`, process_values in pmix_client_get.c:724-731) — `free()` of a non-allocation-start pointer, i.e. heap corruption/SIGABRT. Suggest scanning all `info.len` entries for the directive key (cheap loop) to match upstream semantics.

src/data_ops/mod.rs
```diff
@@ -252,20 +254,37 @@ extern "C" fn get_value_callback_bridge(
...
-            _not_thread_safe: std::marker::PhantomData,
-        })
+        if qualified {
+            // SAFETY: `kv` is valid during the callback; Value_xfer does not modify it.
+            let mut copied = unsafe { std::mem::zeroed() };
+            let copy_status = crate::pmix_ffi_or_mock!(
+                mock = unsafe { mock_ffi::mock_value_xfer(&mut copied, kv) },
+                real = unsafe { ffi::PMIx_Value_xfer(&mut copied, kv) },
+            );
+            (copy_status == ffi::PMIX_SUCCESS as i32).then_some(PmixOwnedValue {
```
SedahsReview, line 280: **Warning (leak):** on the server round-trip route, `_getnb_cbfunc` hands the callback a *fresh* heap struct (`PMIX_VALUE_CREATE` + `PMIx_Value_xfer`, pmix_client_get.c:693-694) that OpenPMIx never frees (`cbdes` does not release `cb->value`). This branch xfers FROM `kv` and never reclaims it, so each qualified get_nb on that route leaks struct + payloads — the exact leak class #389 targets. The bridge cannot safely `Box::from_raw(kv)` here because the local-GDS route delivers `&iptr[0].value` (into C memory), so there is no clean fix; at minimum add a comment documenting the residual leak on the round-trip route.

src/value.rs
```diff
@@ -570,9 +570,9 @@ fn write_payload(dst: &mut pmix_value, payload: PmixPayload) {
 /// After this call `v` is zeroed and safe to drop or reuse.
 ///
 /// # Safety
-/// `v` must have been produced by this crate's builder (or have equivalent
-/// allocation discipline).  Calling this on a `pmix_value_t` produced by the C
-/// library (and therefore managed by `PMIX_VALUE_RELEASE`) is a double-free.
+/// `v` must have been produced by this crate's builder, or be a Rust-owned copy
```
SedahsReview, line 573: **Suggestion:** this doc update landed in dead code — `src/value.rs` is not declared as a module in lib.rs (no `mod value;`, no `#[path]`), and the compiled `PmixOwnedValue`/`free_value` live in lib.rs (lines ~2900/2970; note value.rs's `PmixOwnedValue` lacks the `_not_thread_safe` field the bridge constructs with). The LIVE `free_value` safety doc at lib.rs:2897-2899 still says a C-library-produced value "is a double-free" — now misleading, since the new blocking/plain-nb paths feed PMIx-produced values into `free_value` via the reclaimed-copy pattern. Please apply the doc change to lib.rs (and consider pruning value.rs).

SedahsReview (changes requested):
## Review verdict: REQUEST_CHANGES

The core ownership handoff is correct on all three paths and the full suite passes, but one memory-safety edge in the qualified-directive detection should be fixed before merge (see inline comment on `get_nb`).

### Ownership analysis (verified against OpenPMIx 6.1.0 source, not just headers)
- **Blocking `get_value`** — sound. Upstream `PMIx_Get` hands over the struct (`*val = cb->value; cb->value = NULL;` pmix_client_get.c:371-372, from `PMIX_VALUE_CREATE` calloc). `Box::from_raw(value)` reclaims only the struct (no Drop impl on `pmix_value_t`); the Rust copy's `free_value` frees nested payloads exactly once. Error paths never touch `value`. The added `!value.is_null()` guard also removes a latent null-deref UB in the old code.
- **Plain `get_nb`** — sound on both upstream delivery routes. The `pmix_cb_t` destructor (`cbdes`, pmix_globals.c:465-481) does **not** release `p->value`; `kv->value` is NULL-protected before the kval is released (pmix_client_get.c:697-699). `ptr::read` + `Box::from_raw(kv)` frees the struct once, payloads once.
- **Qualified `get_nb`** — no write to C-owned memory, no double-free. `PMIx_Value_xfer` is a documented COPY (pmix.h:1782) that allocates fresh payloads in `dest`; the source is untouched. On xfer failure, `then_some` returns `None` and the zeroed `copied` drops as a plain C struct (no Drop impl) — safe. Note: when the value arrives via the server round-trip, upstream already hands the callback a fresh heap struct (pmix_client_get.c:693-694) which PMIx never frees, so that route leaks struct+payloads per call; no safe reclaim exists since the bridge cannot distinguish that route from the local route where `kv` points into a C-owned info array (`&iptr[0].value`, pmix_client_get.c:724-731). Worth documenting.

### QUALIFIED_GETS lifecycle
Consulted before value extraction; removed exactly once (bridge fires once per op); req_ids are monotonic so stale entries can never misclassify. Two minor notes: the immediate-failure branch in `get_nb` removes the callback but not the set entry (tiny leak, harmless), and a callback that never fires leaks its entry (same pre-existing pattern as GET_REGISTRY).

### Quality gates (PMIX_PREFIX=$HOME/.local/openpmix-6.1.0)
- `cargo build` — clean.
- `cargo test --lib -- --test-threads=1` — **1828 passed, 0 failed, 29 ignored** (re-verified on a forced rebuild of the changed sources).
- `cargo clippy --lib -- -D warnings` — 6 errors, **all pre-existing** `missing_safety_doc` lints in generated `target/.../out/bindings.rs` (bindgen output); none in src/ files changed by this PR.
- `cargo fmt --check` — 588 pre-existing repo-wide diffs; **none inside this PR's hunks** (verified by region).

### Other findings
- `src/value.rs:573-575` doc update landed in **dead code** — `value.rs` is not declared as a module in lib.rs; the live `free_value` doc (lib.rs:2897-2899) still claims a C-library-produced value is a double-free, which the new reclaim pattern now contradicts. Move the doc change to lib.rs.
- `mock_value_xfer` (mock_ffi.rs:1903) is a no-op that returns success without copying; the qualified branch has **zero test coverage** — the two fixed tests only cover the plain path. A qualified-path bridge test would be valuable.
- PR body note: "test compilation issues" — the full `cargo test --lib` compiles and passes cleanly on this head.

SedahsReview:
## Code Review Summary — PR #395

**Verdict: 🔴 Changes Requested** (1 blocking, 2 warnings, 2 suggestions) — formal review `#4889517924` posted with 3 inline comments.

### 🔴 Critical
- **src/data_ops/mod.rs:365-378** — qualified-directive detection only inspects `info[0]`. OpenPMIx treats info directives position-independently (the array is forwarded whole; `process_request` scans all entries for the directives it handles, and `PMIX_QUALIFIED_VALUE` is passed through unprocessed). A caller placing the directive at index > 0 (e.g. `InfoBuilder` with `external_progress(true)` before `add(b"pmix.qual.val", ...)`) gets a qualified response from PMIx, but the bridge takes the **plain path** → `Box::from_raw(kv)` on `&iptr[0].value`, which points INTO the C-owned info array (`process_values`, pmix_client_get.c:724-731) → `free()` of a non-allocation-start pointer → heap corruption/SIGABRT. Fix: loop over all `info.len` entries (3 lines) to match upstream semantics.

### ⚠️ Warnings
- **src/data_ops/mod.rs:267-277** — qualified branch leaks the original `kv` on the server round-trip route: `_getnb_cbfunc` hands the callback a *fresh* heap struct (`PMIX_VALUE_CREATE` + `PMIx_Value_xfer`, pmix_client_get.c:693-694) that OpenPMIx never frees (`cbdes`, pmix_globals.c:465-481, does not touch `cb->value`). Per-call leak of struct + payloads on that route; no safe reclaim exists (can't distinguish from the local route where `kv` points into C memory) — at minimum document it.
- **src/value.rs:573-575** — doc update landed in **dead code**: `value.rs` is not compiled (no `mod value;` in lib.rs; live `PmixOwnedValue`/`free_value` are lib.rs:2970/2900). The live `free_value` safety doc (lib.rs:2897-2899) still claims a C-library-produced value is a double-free — now contradicted by the new reclaim pattern. Move the doc change to lib.rs.

### 💡 Suggestions
- `get_nb` immediate-failure branch (mod.rs:409-415) removes the callback from GET_REGISTRY but not the req_id from QUALIFIED_GETS → stale set entry per failed call (harmless: monotonic req_ids never reused).
- `mock_value_xfer` (mock_ffi.rs:1903) is a no-op returning success without copying, and the qualified branch has **zero test coverage** — the two fixed tests cover only the plain path. A qualified-path bridge test would be valuable.

### ✅ Looks Good (ownership, verified against OpenPMIx 6.1.0 source)
- **Blocking `get_value`**: struct reclaimed exactly once (`*val = cb->value; cb->value = NULL;` pmix_client_get.c:371-372, from `PMIX_VALUE_CREATE`); copy's `free_value` frees nested payloads exactly once; `!value.is_null()` guard also removes a latent null-deref UB. No leak/double-free on any path.
- **Plain `get_nb`**: sound on both upstream routes — `cbdes` does not release `cb->value`, `kv->value` NULL-protected before kval release (pmix_client_get.c:697-699). `ptr::read` + `Box::from_raw` frees struct once, payloads once.
- **Qualified `get_nb`**: `PMIx_Value_xfer` is a documented COPY (pmix.h:1782 — fresh payloads, source untouched); failure path drops a zeroed struct safely (`then_some` — no `PmixOwnedValue` constructed); no write to C-owned memory.
- **QUALIFIED_GETS**: consulted before extraction, removed exactly once, monotonic req_ids prevent stale misclassification.
- **CStr check sound**: `Info` is crate-constructed via `PMIx_Info_load` (NUL-terminated keys); `pmix.qual.val` = 13 bytes fits `pmix_key_t[512]`.
- Mock integration balanced (`mock_get` `Box::into_raw` ↔ blocking `Box::from_raw`); the two fixed tests heap-allocate correctly; no other stack-memory bridge callers.
- All `unsafe` blocks carry `// SAFETY:`; no stale zeroing-hack comments, conflict markers, secrets, or debug leftovers.

### Quality gates (PMIX_PREFIX=$HOME/.local/openpmix-6.1.0)
- `cargo build` — ✅ clean
- `cargo test --lib -- --test-threads=1` — ✅ **1828 passed, 0 failed, 29 ignored** (re-verified on a forced rebuild of the changed sources)
- `cargo clippy --lib -- -D warnings` — ⚠️ 6 errors, **all pre-existing** `missing_safety_doc` in generated `bindings.rs` (bindgen output); **none** in src/ files changed by this PR
- `cargo fmt --check` — ⚠️ 588 pre-existing repo-wide diffs; **none** inside this PR's hunks

---
*Reviewed by Hermes Agent (code-reviewer). Note: the PR body's "test compilation issues" caveat is stale — the full `cargo test --lib` compiles and passes cleanly on head `29bd1ab`.*

src/data_ops/mod.rs
```diff
@@ -343,6 +366,24 @@ pub fn get_nb(
...
+            let entries = unsafe { std::slice::from_raw_parts(info.handle, info.len) };
+            let qualified = entries.iter().any(|entry| {
+                // SAFETY: each entry's key is a NUL-terminated pmix_key_t.
+                unsafe { CStr::from_ptr(entry.key.as_ptr()).to_bytes() == b"pmix.qual.val" }
+            });
+            if qualified {
+                QUALIFIED_GETS
+                    .lock()
+                    .expect("mutex poisoned (data_ops.rs)")
+                    .insert(req_id);
```
SedahsReview, line 384: **Warning (new): `QUALIFIED_GETS` entry leaks when `PMIx_Get_nb` fails synchronously.** The scan inserts `req_id` here before the FFI call, but the immediate-failure branch (~lines 416-422) removes only the `GET_REGISTRY` entry. Because the callback — which does `.remove(&req_id)` (~line 260) — is never invoked on that path (per the function's own doc: "The callback will NOT be called"), the set entry is never reclaimed. A long-lived process issuing failing qualified gets (e.g. after a PMIx disconnect, bad proc) grows the set without bound — the exact leak class this PR exists to fix. Fix: add `QUALIFIED_GETS.lock().expect("mutex poisoned (data_ops.rs)").remove(&req_id);` next to the registry cleanup in the `else` branch.

src/data_ops/tests.rs
```diff
@@ -2060,15 +2060,19 @@ use super::*;
             registry.insert(req_id, Box::new(TestGetCb));
         }
 
-        // Create a mock pmix_value_t for the callback
-        let mut mock_value: ffi::pmix_value_t = unsafe { std::mem::zeroed() };
+        // Create a heap-allocated mock pmix_value_t for the callback.
+        // SAFETY: pmix_value_t is a C representation and zero is a valid
+        // initialization state before setting the type field.
```
SedahsReview, line 2065: **Suggestion (testing gap):** the qualified path has no test coverage — neither the all-entries directive scan (the fix for the previous blocking issue, which only fires with the directive at index > 0) nor the xfer branch in the callback. The mock path can exercise it: seed `QUALIFIED_GETS` directly (tests live in the same module) and drive `get_value_callback_bridge` with a heap value, asserting the xfer branch is taken and `kv` is not reclaimed; or extend `mock_get_nb` to honor info directives. A regression test would have caught the original info[0]-only bug.

SedahsReview (changes requested):
## Re-review verdict: REQUEST_CHANGES

All three prior findings are verified resolved (below), but this round surfaces one new warning-level issue in the new `QUALIFIED_GETS` machinery: the request id is never removed from the set when `PMIx_Get_nb` fails synchronously, so repeated failing qualified gets leak one set entry each. One-line fix — see inline comment.

### Prior findings — all resolved
1. **BLOCKING (info[0]-only qualified detection) — FIXED.** `get_nb` now scans every entry via `std::slice::from_raw_parts(info.handle, info.len)` + `iter().any(...)`. Verified sound: `Info` is only constructed via `InfoBuilder::build` / `info_with_string_key`, both through `PMIx_Info_create` + `PMIx_Info_load` (NUL-terminated `pmix_key_t`, PMIX_MAX_KEYLEN=511), so each `CStr::from_ptr` read stays in-bounds; `// SAFETY:` comments are present on both unsafe ops; null handle / len 0 are skipped and an empty key yields empty bytes — no panic, no false positive.
2. **WARNING (residual leak on server round-trip route) — documented.** The comment at src/data_ops/mod.rs:268-271 accurately describes that the xfer leaves the fresh heap struct + nested payloads leaked and explains why it cannot be reclaimed (local-GDS route delivers a pointer into C-owned info memory). No overclaim.
3. **SUGGESTION (doc in dead code) — applied to live code.** src/lib.rs:2897-2899 now carries the Rust-owned-copy wording. The identical change remains in dead `src/value.rs` (not a module in lib.rs, never compiled) — harmless; consider pruning value.rs, non-blocking.

### Previously-approved ownership analysis — re-verified, still sound
- Blocking `get_value`: `*value` copy + `Box::from_raw` reclaim; the added `!value.is_null()` guard is a good defensive fix.
- Plain `get_nb`: `ptr::read(kv)` + `Box::from_raw(kv)`; heap-allocated test values (Box::into_raw) now mirror the real callback contract.
- Qualified `get_nb`: xfer into a fresh zeroed struct; no write to C-owned memory; only the documented residual leak.
- No double-free, no write to C-owned memory on any path.

### Quality gates
- `cargo build` — clean
- `cargo test --lib -- --test-threads=1` — **1828 passed, 0 failed, 29 ignored**
- `cargo clippy --lib -- -D warnings` — 6 errors, all pre-existing `missing_safety_doc` in generated bindings (target/debug/build/pmix-*/out/bindings.rs); none in src/ files touched by this PR
- `cargo fmt --check` — pre-existing repo-wide baseline (the 4 touched files already differ from rustfmt at base `main`); PR-added lines are consistent with surrounding style

SedahsReview:
## Re-review summary — PR #395 (head 6401809)

**Verdict: REQUEST_CHANGES** (1 new warning; all 3 prior findings resolved). Review: https://github.com/SedahsDev/pmix-rs/pull/395#pullrequestreview-4889536928

### Critical
- None.

### Warnings
- **src/data_ops/mod.rs:416-422 (new)** — `QUALIFIED_GETS` entry leaks when `PMIx_Get_nb` fails synchronously. The qualified scan inserts `req_id` before the FFI call, but the immediate-failure branch only removes the `GET_REGISTRY` entry; the callback (which removes the qualified entry) is never invoked on that path. Repeated failing qualified gets grow the set without bound. One-line fix: `QUALIFIED_GETS.lock().expect("mutex poisoned (data_ops.rs)").remove(&req_id);` in the `else` branch. (Inline comment at mod.rs:382.)

### Suggestions
- **src/data_ops/tests.rs** — No test covers the qualified path (all-entries scan with the directive at index > 0, nor the xfer branch). A regression test would have caught the original info[0]-only bug. (Inline comment at tests.rs:2065.)
- **src/value.rs:573** — The doc change remains in dead code (`src/value.rs` is not a module in lib.rs, never compiled). Harmless; consider pruning value.rs. The LIVE doc at src/lib.rs:2897-2899 is correctly updated.

### Prior findings — verified resolved
1. **BLOCKING (info[0]-only qualified detection) — FIXED.** `get_nb` now scans all entries via `std::slice::from_raw_parts(info.handle, info.len)` + `iter().any(..)`. Slice is sound (`Info` is only built via `PMIx_Info_create`/`PMIx_Info_load`, keys NUL-terminated at PMIX_MAX_KEYLEN=511); `// SAFETY:` comments present on both unsafe ops; null handle / len 0 / empty key all handled without panic or false positive.
2. **WARNING (residual round-trip leak) — documented.** Comment at src/data_ops/mod.rs:268-271 is accurate and does not overclaim.
3. **SUGGESTION (dead-code doc) — applied to the live `free_value`** at src/lib.rs:2897-2899 (Rust-owned-copy wording).

### Looks Good
- Ownership handoff re-verified on all three paths: blocking (copy + `Box::from_raw` reclaim, plus a new defensive `!value.is_null()` guard), plain get_nb (`ptr::read` + `Box::from_raw`), qualified get_nb (xfer into fresh zeroed struct; C-owned info memory never written). No double-free on any path.
- Test values correctly heap-allocated (`Box::into_raw`) to mirror the real `PMIx_Get_nb` callback contract.
- `cargo build` — clean.

### Quality gates
- `cargo test --lib -- --test-threads=1` — **1828 passed, 0 failed, 29 ignored** (full suite)
- `cargo clippy --lib -- -D warnings` — 6 errors, ALL pre-existing `missing_safety_doc` in generated bindings (`target/debug/build/pmix-*/out/bindings.rs`, present on `main` too); none in src/ files touched by this PR
- `cargo fmt --check` — pre-existing repo-wide baseline; the 4 touched files already differ from rustfmt at base `main`; PR-added lines are consistent with surrounding style

src/data_ops/mod.rs
```diff
@@ -375,6 +416,12 @@ pub fn get_nb(
     } else {
         // Immediate failure — remove the registered callback so it
         // will never be invoked.
+        // The completion callback never runs on this path, so drop both the
+        // callback registration and any qualified-value marker.
+        QUALIFIED_GETS
```
SedahsReview, line 423: 💡 Suggestion (non-blocking): the fix is correct, but there's no regression test exercising it. `mock_ffi` supports per-function status overrides (FUNCTION_OVERRIDES / MockGuard), so a test could set `PMIx_Get_nb` to fail, call `get_nb` with a `pmix.qual.val` info directive, and assert `QUALIFIED_GETS` is empty afterward. Same for the qualified success path (xfer branch) — currently untested.

src/data_ops/mod.rs
```diff
@@ -252,20 +254,41 @@ extern "C" fn get_value_callback_bridge(
         None => return, // Callback already consumed or never registered.
     };
 
+    let qualified = QUALIFIED_GETS
```
SedahsReview, line 259: 💡 Suggestion (non-blocking): the null-cbdata early return above (line 238-240) skips this removal and would leak both registries — but it's unreachable: `encode_req_id` guarantees non-null (req_id starts at 1; PMIx returns cbdata verbatim). Also note src/value.rs's free_value doc was edited but value.rs is an orphan file not registered in lib.rs (never compiled) — harmless, but the live doc edit in lib.rs is the one that matters.

SedahsReview (approved):
## Round-3 review verdict: APPROVE

Re-reviewed after the one-line error-path fix (315c7fb). The round-2 warning is resolved and all previously approved items still hold.

### Error-path leak — FIXED (verified)
`get_nb`'s immediate-failure branch now runs `QUALIFIED_GETS.lock()...remove(&req_id)` (src/data_ops/mod.rs:421-424) before the `GET_REGISTRY` cleanup. `.remove()` on an absent key is a no-op, so the unconditional removal is safe for non-qualified requests. Leak paths audited:
- Success path → async callback removes the qualified marker at bridge top (mod.rs:257-260).
- Immediate failure → removed here now.
- `CString::new(key)` failure → returns before the qualified marker is ever inserted.
- Bridge `None => return` (callback already consumed/never registered) → the qualified entry was already removed on the consuming invocation; insert order (registry then qualified, same call) guarantees no dangling qualified-only entry.
- Lock ordering: all three mutexes are taken scoped and never nested across sections (bridge: GET_REGISTRY then QUALIFIED_GETS, both dropped before user callback; error path: QUALIFIED_GETS then GET_REGISTRY, sequential). No deadlock window.

### Null-cbdata residual — unreachable, non-blocking
`get_value_callback_bridge`'s `cbdata.is_null()` early return (mod.rs:238-240) precedes both registry removals, so a null-cbdata invocation would leak both entries. This is unreachable in practice: `encode_req_id` requires `req_id != 0` (debug_assert + GET_SEQ starts at 1), and PMIx passes back exactly the cbdata we supplied. Residual nit only.

### Re-confirmed (round 1/2 items)
- All-entries qualified scan: `slice::from_raw_parts(handle, len).iter().any(...)` — scans every info entry, not just the first.
- Residual round-trip leak of the qualified-value struct is documented at mod.rs:268-271 with the local-GDS vs server-route rationale.
- `free_value` doc in live lib.rs updated (both copies; value.rs copy is dead code — see suggestion).
- Heap-allocated mock test values (`Box::into_raw`) match the bridge's new `Box::from_raw` ownership contract.
- No C-memory writes remain (old `ptr::write(kv, zeroed())` removed); non-qualified path reads the struct then frees the allocation via `Box::from_raw`.
- No double-free: verified against OpenPMIx source (pmix_client_get.c / pmix_globals.c): the `pmix_cb_t` destructor does NOT release `cb->value`, and the GDS hash/shmem3 fetch paths hand back freshly malloc'd/calloc'd value structs — the value is abandoned to the callback, so reclaiming it is exactly-once.

### Quality gates (all pass)
- `cargo build` — clean
- `cargo test --lib --features mock_ffi -- --test-threads=1` — **1828 passed, 0 failed**
- `cargo clippy --lib --features mock_ffi -- -D warnings` — only pre-existing errors (bindgen missing_safety_doc, unnecessary-unsafe in mock macro branches, etc.); zero in PR-touched lines (verified vs main)
- `cargo fmt --check` — only pre-existing diffs; PR-touched files show no new fmt issues (branch has fewer flagged lines than main in these files)

Two non-blocking suggestions in inline comments. Approving.

SedahsReview:
## Round-3 review summary — APPROVED ✅

**Reviewed commit:** 315c7fb0141d7119ba4af4fe196a80d03c1bd0fc (`fix/value-ownership-handoff`)

### Verdict
**APPROVE** — the round-2 warning (QUALIFIED_GETS entry leak on synchronous `PMIx_Get_nb` failure) is fixed, and no new critical/warning issues appeared. Round-1 and round-2 findings are all resolved.

### Critical
None.

### Warnings
None. The round-2 warning is confirmed fixed:
- `get_nb` immediate-failure branch now removes the QUALIFIED_GETS marker (src/data_ops/mod.rs:421-424) before `GET_REGISTRY` cleanup; `.remove()` on an absent key is a no-op, so non-qualified requests are unaffected.
- All other leak paths audited clean: success → callback removes at bridge top; `CString::new` failure → returns before marker insertion; bridge `None => return` → marker already consumed or never inserted (insert order guarantees no dangling qualified-only entry).
- No lock-ordering hazard: all three mutexes scoped, never nested.

1. **mod.rs:421** — Add a regression test: use mock_ffi's per-function status override to force `PMIx_Get_nb` failure with a `pmix.qual.val` directive, then assert `QUALIFIED_GETS` is empty. The qualified success (xfer) path is also untested.
2. **mod.rs:257** — The null-cbdata early return (line 238-240) precedes this removal and would leak both registries if hit; it is unreachable in practice (`encode_req_id` guarantees non-null cbdata, PMIx returns cbdata verbatim), so this is a theoretical nit only. Also: the src/value.rs free_value doc edit touches an orphan file not registered in lib.rs (never compiled) — harmless, but the lib.rs copy is the live one.

### Looks Good
- All-entries qualified scan via `slice::from_raw_parts` + `iter().any`.
- Residual qualified round-trip leak documented with local-GDS vs server-route rationale (mod.rs:268-271).
- Heap-allocated mock test values match the bridge's `Box::from_raw` ownership contract.
- No C-memory writes (old `ptr::write(kv, zeroed())` removed); no double-free — verified against OpenPMIx source that the value struct is abandoned to the callback (cb destructor does not release `cb->value`).
- `free_value` doc in live lib.rs updated.

### Quality gates (head 315c7fb, mock-FFI env)
- `cargo build` — clean ✅
- `cargo test --lib --features mock_ffi -- --test-threads=1` — **1828 passed, 0 failed** ✅
- `cargo clippy --lib --features mock_ffi -- -D warnings` — only pre-existing errors (bindgen missing_safety_doc, unnecessary-unsafe in mock macro branches); zero in PR-touched lines (verified vs main) ✅
- `cargo fmt --check` — only pre-existing diffs; no new issues in PR-touched files ✅

---
*Reviewed by SedahsReview (Hermes code-reviewer) — round 3*

PR #396, opened 2026-08-08
perf(threading): back spawn_from_callback with a bounded worker pool

### LINKED ISSUE #388: perf: spawn_from_callback creates a fresh OS thread per callback hop — thread churn, no backpressure
## Summary

`threading::spawn_from_callback` (`src/threading.rs:157-177`) spawns a **new OS thread per invocation** (`std::thread::Builder::new().name("pmix-callback-hop")`). Every `_nb` completion, event notification, or server upcall that uses the documented hop pattern creates a thread (~8 MiB virtual stack reservation each).

## Impact

- High-frequency paths — IOF streams, heartbeat/event storms, many concurrent `_nb` completions — churn threads and can exhaust `RLIMIT_NPROC` or slow the process with scheduler pressure.
- The function is the crate's recommended escape hatch for "must not block the progress thread", so it will be on the hot path in real MT apps.
- No pooling, no concurrency limit, no backpressure: a burst of completions spawns a burst of threads.

## Suggested fix

- Keep the fire-and-forget API, but back it with a small shared worker pool (bounded queue + N workers, sized to `available_parallelism`) with fallback to direct spawn when the queue is full, or
- Document and provide a `CallbackChannel`-first pattern for hot paths (already exists, but `spawn_from_callback` is the path of least resistance), or
- At minimum, make the thread name/stack size configurable and document the per-hop cost.

The API contract (never join in-handler, panics confined to the hop thread) is unchanged.

SedahsDev:
Closes #388

## Summary
- Replaces per-callback thread creation with a process-global detached worker pool.
- Uses one worker per available parallelism (falling back to four) and a bounded queue with four slots per worker.
- Uses non-blocking enqueue; a full or disconnected queue falls back to a dedicated `pmix-callback-hop` thread so work is never dropped and the PMIx progress thread never waits.
- Pool workers are detached and need no finalize-time shutdown.

## API changes
- `spawn_from_callback` now returns `CallbackJoinHandle` instead of `std::thread::JoinHandle<()>`.
- `CallbackJoinHandle::join` preserves completion waiting and panic reporting for pooled and fallback work.
- All in-repository callers already use the returned value as fire-and-forget; no caller changes were required.

## Test plan
- `PMIX_PREFIX=$HOME/.local/openpmix-6.1.0 cargo build` (pass)
- `PMIX_PREFIX=$HOME/.local/openpmix-6.1.0 cargo test` (pass; daemon tests remain ignored without PRTE/DVM)
- Targeted burst, fallback, and panic-containment tests (pass)
- `cargo fmt --check` (pass before final targeted test; later full-suite build formatting exposed pre-existing unrelated formatting drift)
- `cargo clippy --all-targets -- -D warnings` (fails on pre-existing repository warnings/errors outside this change; changed code had only a fixable `question_mark` lint, fixed before publication)

### DIFF
--- a/src/threading.rs
+++ b/src/threading.rs
@@ -41,8 +41,8 @@
 //! The two primitives cover the two ways a handler can get work off the
 //! progress thread:
 //!
-//! * [`spawn_from_callback`] — fire-and-forget: start a fresh application
-//!   thread for blocking work and return immediately.
+//! * [`spawn_from_callback`] — fire-and-forget: queue blocking work on the
+//!   shared callback worker pool and return immediately.
 //! * [`CallbackChannel`] — request/response: the application thread owns the
 //!   channel (receiver side) and processes messages the handler sends.
 //!
@@ -57,7 +57,7 @@
 //! // Inside a PMIx callback (progress thread):
 //! let _ctx = ProgressContext;                 // documents: we are on progress
 //! let _ = tx.send((0, b"payload".to_vec()));  // cheap, non-blocking
-//! let _ = spawn_from_callback(move || {       // or: fresh thread for blocking work
+//! let _ = spawn_from_callback(move || {       // or: pooled thread for blocking work
 //!     // blocking PMIx calls are legal here
 //! });
 //!
@@ -75,7 +75,9 @@
 //! `examples/server_upcall_hop.rs` (server module fence/modex upcalls +
 //! delayed `cbfunc`), plus [THREADING.md](../THREADING.md).
 
+use std::any::Any;
 use std::sync::mpsc::{self, Receiver, RecvError, RecvTimeoutError, Sender, TryRecvError};
+use std::sync::{Arc, Mutex, OnceLock};
 
 /// Zero-sized marker for code that runs on the PMIx **progress thread**.
 ///
@@ -122,8 +124,10 @@ pub struct ProgressContext;
 ///
 /// The callback runs on the PMIx progress thread; any work that could block
 /// (including blocking PMIx calls) must be moved to an application thread.
-/// This helper spawns a fresh, named thread to run `f` and returns a
-/// [`JoinHandle`](std::thread::JoinHandle) the caller may store or detach.
+/// This helper queues `f` on a process-wide, bounded pool of named worker
+/// threads and returns a [`CallbackJoinHandle`] the caller may store or
+/// detach. If the queue is full, it falls back to a fresh named thread rather
+/// than blocking the PMIx progress thread or dropping work.
 ///
 /// # Rules
 ///
@@ -140,8 +144,7 @@ pub struct ProgressContext;
 ///   (the default panic hook does not call `exit()` for non-main threads).
 ///   Under `panic = "abort"` the process aborts regardless of this helper.
 ///   This helper logs a branded diagnostic on stderr for fire-and-forget
-///   callers, then resumes unwinding so a joined handle still reports
-///   [`JoinHandle::join`](std::thread::JoinHandle::join) `Err`. Prefer not
+///   callers; a joined [`CallbackJoinHandle`] still reports `Err`. Prefer not
 ///   panicking in hop work in long-running HPC jobs — treat panics as bugs.
 ///
 /// # Example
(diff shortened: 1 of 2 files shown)


src/threading.rs
```diff
@@ -154,26 +157,134 @@ pub struct ProgressContext;
...
-            if let Err(payload) = std::panic::catch_unwind(std::panic::AssertUnwindSafe(f)) {
-                eprintln!(
-                    "pmix: thread 'pmix-callback-hop' panicked while running \
-                     work hopped off the PMIx progress thread"
-                );
-                std::panic::resume_unwind(payload);
-            }
+    let pool = match CALLBACK_POOL.get_or_init(CallbackPool::new) {
+        Ok(pool) => pool,
+        Err(error) => return Err(std::io::Error::new(error.kind(), error.to_string())),
```
SedahsReview: **Warning (non-blocking):** `OnceLock` caches a failed pool init forever. If `CallbackPool::new()` ever fails (e.g. a transient thread-spawn `EAGAIN` while creating workers), every subsequent `spawn_from_callback` returns `Err` and the direct-spawn fallback is **never attempted again** for the process lifetime — the old code retried per call. Since callers in PMIx callbacks treat `Err` as "log and continue", hopped work silently never runs after that point, which can reintroduce exactly the progress-thread blocking this helper exists to prevent. Suggest falling through to the direct-spawn path on init failure (consistent with the `Full`/`Disconnected` fallback), or at minimum documenting the permanent-disable semantics. Also note `error.kind()` + `error.to_string()` drops the original OS error detail (kind is preserved).

src/threading.rs
```diff
@@ -154,26 +157,134 @@ pub struct ProgressContext;
...
+                completion: completion_rx,
+                direct: Some(direct),
+            })
+        }
+    }
+}
+
+const CALLBACK_QUEUE_CAPACITY_MULTIPLIER: usize = 4;
+
+type CallbackResult = Result<(), Box<dyn Any + Send + 'static>>;
```
SedahsReview: **Suggestion (cosmetic):** `CallbackResult` is a private alias returned by the public `CallbackJoinHandle::join` (line 277). It compiles clean because aliases are transparent and the expanded type (`Result<(), Box<dyn Any + Send + 'static>>` == `std::thread::Result<()>`) is fully public — downstream joiners can spell it as `std::thread::Result<()>` — but the crate's own return type is unnameable. Consider making the alias `pub` since it is already part of the public surface through `join`'s signature.

src/threading.rs
```diff
@@ -451,4 +559,30 @@ mod tests {
...
+                    let _ = release_rx.lock().unwrap().recv();
+                    completed.fetch_add(1, Ordering::SeqCst);
+                })
+                .expect("submit backpressure task"),
+            );
+        }
+        for _ in 0..100 {
+            release_tx.send(()).expect("release task");
+        }
+        for handle in handles {
```
SedahsReview, line 606: **Suggestion (test robustness):** the join loop has no timeout. If a future change ever drops work, `handle.join()` blocks forever and CI hangs instead of failing fast — a bounded wait would make a regression visible. Not an issue in the current code: the 100 releases exactly balance the 100 one-shot `recv`s, and every join returns only after its closure ran.

SedahsReview (approved):
## Review: APPROVE

Verified head ef12fc9 locally against main. The bounded worker pool is sound: panic containment, backpressure fallback, and join semantics all check out. One non-blocking warning and two suggestions in the inline comments.

### Panic semantics (verified)
- **Pooled task**: `run_callback_task` catches the panic inside the worker loop iteration, logs the branded diagnostic, sends `Err(payload)` on the completion channel, and does **not** resume — the worker survives and keeps recv()ing. `join()` surfaces `Err(payload)` via `completion.recv()`.
- **Direct-fallback task**: catch_unwind -> branded log -> `resume_unwind`. The `send(Ok(()))` after the `if let Err` is unreachable when the closure panicked (`resume_unwind` diverges), so `direct.join()?` propagates `Err(payload)` before `completion.recv()` is ever reached. No path both sends and resumes; no swallowed panic; no double-report (`resume_unwind` does not re-run the panic hook); the completion channel is fully owned by task + handle and drops cleanly — no dangling channel.
- `direct.join()?` type-checks because `CallbackResult` is structurally identical to `std::thread::Result<()>` (confirmed by compilation).

### Backpressure test (deadlock analysis)
100 tasks each do exactly one `release_rx.recv()` before completing; the main thread sends 100 releases on an **unbounded** channel — exactly balanced, no deadlock. `try_send` never blocks; `Full` and `Disconnected` both fall back to a direct spawn carrying the returned task, so work is never dropped and the progress thread never blocks. Queue capacity (workers*4) < 100 guarantees the fallback path is exercised, and every `join()` returns only after its closure ran, so `completed == 100` is deterministic.

### Worker lifecycle
Workers break on channel disconnect; the pool is a process-global `OnceLock`, worker `JoinHandle`s are dropped (detached), so no finalize-time shutdown is needed — documented in THREADING.md. Detached std threads do not block process exit.

### API change assessment
Return type `JoinHandle<()>` -> `CallbackJoinHandle` is breaking but intentional and documented (PR body + THREADING.md + module docs), and permitted by AGENTS.md (no backward-compat preservation). All in-repo callers are fire-and-forget (drop = detach); joiners observe identical semantics (`Err(payload)` on panic). Note: **there was no public `CallbackResult` on main** — the new private alias is freshly introduced (see inline suggestion).

### Quality gates (PMIX_PREFIX=$HOME/.local/openpmix-6.1.0)
- `cargo build`: PASS
- `cargo test --lib threading::tests -- --test-threads=1`: 12/12 PASS
- `cargo test --lib -- --test-threads=1`: 1831 passed, 0 failed, 29 ignored
- `cargo clippy --lib -- -D warnings`: only the 6 pre-existing bindgen `missing_safety_doc` errors in generated `target/.../out/bindings.rs`; **zero findings in src/threading.rs**
- `cargo fmt --check`: src/threading.rs clean under edition 2024 (repo-wide drift elsewhere is pre-existing on main)
- `cargo check --features mock_ffi --lib`: PASS (44 baseline warnings, 0 errors)

No `unsafe` introduced. Solid work.

SedahsReview:
## Code Review Summary — PR #396 (perf(threading): bounded worker pool)

**Verdict: APPROVE** (1 non-blocking warning, 2 suggestions) — reviewed commit ef12fc9 on branch perf/callback-hop-worker-pool.

### Critical
- None.

### Warnings
- **src/threading.rs:166** — `OnceLock` permanently caches a failed pool init: one transient thread-spawn failure disables `spawn_from_callback` (and its direct-spawn fallback) for the whole process lifetime, whereas the old code retried per call. Since callback-side callers treat `Err` as "log and continue", hopped work can silently stop happening, reintroducing progress-thread blocking. Suggest falling through to the direct-spawn path on init failure, or documenting the permanent-disable semantics. Also `error.to_string()` loses OS error detail.

### Suggestions
- **src/threading.rs:203** — private `type CallbackResult` leaks through the public `join()` signature; make the alias `pub` for nameability (cosmetic; it is structurally `std::thread::Result<()>`, so joiners keep identical semantics).
- **src/threading.rs:583** — backpressure test joins lack a timeout; a future work-dropping regression would hang CI rather than fail fast.

### Looks Good
- Panic semantics verified on both paths: pooled → contained (worker survives), branded log, `Err(payload)` via completion; direct → branded log + `resume_unwind`, `Err` via `direct.join()?`. No double-report, no swallowed panic, no dangling channel.
- `try_send` never blocks; `Full`/`Disconnected` both fall back to a direct spawn with the returned task — work never dropped, progress thread never blocked.
- Backpressure test deterministic and deadlock-free (100 one-shot recvs exactly balanced by 100 unbounded sends; every join returns only after its closure ran).
- Worker shutdown: break on disconnect, detached, no finalize-time shutdown — documented in THREADING.md.
- API change (`JoinHandle<()>` → `CallbackJoinHandle`) breaking but intentional, documented (PR body + THREADING.md + module docs), permitted by AGENTS.md. Note: no public `CallbackResult` existed on main — the new private alias is fresh.
- `assert_ne!` softening correct and non-flaky (test caller is the harness thread, never a pool worker). No `unsafe` introduced.

### Quality gates (PMIX_PREFIX=$HOME/.local/openpmix-6.1.0)
- `cargo build`: PASS
- `cargo test --lib threading::tests -- --test-threads=1`: 12/12 PASS
- `cargo test --lib -- --test-threads=1`: 1831 passed / 0 failed / 29 ignored
- `cargo clippy --lib -- -D warnings`: only 6 pre-existing bindgen `missing_safety_doc` errors (generated bindings.rs); zero findings in src/threading.rs
- `cargo fmt --check`: src/threading.rs clean (edition 2024); repo-wide drift elsewhere pre-existing on main
- `cargo check --features mock_ffi --lib`: PASS (44 baseline warnings, 0 errors)

---
*Reviewed by Hermes Agent (code-reviewer profile)*

PR #397, opened 2026-08-08
perf(client): reduce cached proc read contention

### LINKED ISSUE #387: perf: PmixClientInner.proc: Mutex<Option<Proc>> serializes the hot proc()/rank()/require_proc() path
## Summary

`PmixClientInner` stores the process handle as `Mutex<Option<Proc>>` (`src/lib.rs:3556`). The field is written **only** during `connect` (`:3705`) and cleared in `disconnect` (`:3738`) — every other access is a read. Yet the read-side helpers all take the mutex:

- `proc()` / `get_proc()` — `src/lib.rs:3762-3773`
- `require_proc()` — `:3776-3779` (the README example calls this before every `fence`)
- `rank()` / `get_rank()` — `:3782-3794`
- `proc_with_nspace()` — `:3805-3813`

## Impact

Worker threads that clone `PmixClient` and then call `require_proc()` per operation (the documented MT pattern, `THREADING.md` §3/§7.1) all take the same mutex per call. `pmix_proc_t` is POD (`[c_char; 256]` + rank — `src/proc.rs:47-50`), so the cached handle is a cheap, copyable snapshot; it does not need a `Mutex` for read-mostly access.

## Suggested fix

- `OnceLock<Proc>` for the connected snapshot + `AtomicU8` state gate (the state already exists), or
- `RwLock<Option<Proc>>` so concurrent readers do not serialize, or
- keep the mutex but expose a lock-free `get_proc()` that returns a copied snapshot from a `OnceLock`/`AtomicPtr`-style slot.

The lifecycle contract (one connect/disconnect per process) is unchanged.

SedahsDev:
Closes #387

## Summary
- Replace PmixClientInner.proc: Mutex<Option<Proc>> with RwLock<Option<Proc>>.
- Connect and disconnect retain exclusive write access for lifecycle updates; proc/rank/proc_with_nspace use shared read access, removing reader serialization on the hot path.
- Keep the existing AtomicU8 lifecycle gate and exact public accessor signatures and behavior, including require_proc/require_rank panics and proc_with_nspace returning ErrInit when disconnected.
- Keep PmixClient Send + Sync and update the inner debug description.

## Mechanism
RwLock<Option<Proc>> is the simplest MSRV-compatible read-mostly storage for the existing clear-on-disconnect lifecycle. Shared readers can snapshot the cloneable POD Proc concurrently, while connect/disconnect continue to serialize updates.

## Test plan
- PMIX_PREFIX=$HOME/.local/openpmix-6.1.0 cargo build — passed.
- PMIX_PREFIX=$HOME/.local/openpmix-6.1.0 cargo test — passed.
- Targeted PmixClient unit tests, including the concurrent accessor smoke test — passed.
- cargo clippy --all-targets -- -D warnings — blocked by pre-existing repository-wide lint failures unrelated to this change.
- cargo fmt --check — blocked by pre-existing formatting drift outside the changed file; the staged diff passes git diff --check.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3516,7 +3516,7 @@ pub fn info_with_string_key(key: &str, value: &str) -> Info {
 // ─────────────────────────────────────────────────────────────────────────────
 
 use std::sync::atomic::{AtomicU8, Ordering};
-use std::sync::{Arc, Mutex, OnceLock, RwLock};
+use std::sync::{Arc, OnceLock, RwLock};
 
 /// Lifecycle state of the process-wide PMIx client session.
 ///
@@ -3675,7 +3675,7 @@ impl PmixClient {
     /// Transitions `Uninitialized` → `Live`. Fails if already `Live`
     /// ([`PmixError::ErrExists`]) or after finalize ([`PmixError::ErrInit`]).
     ///
-    /// Serialized on the session mutex so concurrent `connect` calls cannot
+    /// Serialized on the session lock so concurrent `connect` calls cannot
     /// double-init.
     pub fn connect(&self, info: Option<Info>) -> Result<(), PmixError> {
         let mut proc_guard = self
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,7 +13,8 @@
 //!   Use [`PmixClient::connect_new`] / [`connect`](PmixClient::connect) /
 //!   [`disconnect`](PmixClient::disconnect). There is no legacy `Context` / `init`.
 //! - Do **not** lock every put/get in Rust by default — OpenPMIx ≥ 6.1 serializes entry.
-//!   Lock only Rust-owned shared state (the session caches `Proc` under a mutex).
+//!   Lock only Rust-owned shared state (session lifecycle updates use a write lock;
+//!   cached `Proc` reads use shared read locks).
 //! - Data ops stay as free functions ([`put_value`], [`get_value`], [`commit`],
 //!   [`fence`], …). Pass [`Proc`] from the client; build [`Info`] per call.
 //! - [`Info`] and other C-owned handles are **`!Send` + `!Sync`**
@@ -3515,7 +3516,7 @@ pub fn info_with_string_key(key: &str, value: &str) -> Info {
 // ─────────────────────────────────────────────────────────────────────────────
 
 use std::sync::atomic::{AtomicU8, Ordering};
-use std::sync::{Arc, Mutex, OnceLock};
+use std::sync::{Arc, Mutex, OnceLock, RwLock};
 
 /// Lifecycle state of the process-wide PMIx client session.
 ///
@@ -3553,7 +3554,7 @@ impl PmixClientState {
 /// multi-threaded cloning.
 struct PmixClientInner {
     /// Cached proc from `PMIx_Init`; only meaningful while state is `Live`.
-    proc: Mutex<Option<Proc>>,
+    proc: RwLock<Option<Proc>>,
     state: AtomicU8,
 }
 
@@ -3561,7 +3562,7 @@ impl std::fmt::Debug for PmixClientInner {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("PmixClientInner")
             .field("state", &self.state())
-            .field("proc", &"Mutex<Option<Proc>>")
+            .field("proc", &"RwLock<Option<Proc>>")
             .finish()
     }
 }
@@ -3579,7 +3580,7 @@ fn client_session() -> Arc<PmixClientInner> {
     CLIENT_SESSION
         .get_or_init(|| {
             Arc::new(PmixClientInner {
-                proc: Mutex::new(None),
+                proc: RwLock::new(None),
                 state: AtomicU8::new(PmixClientState::Uninitialized as u8),
             })
         })
@@ -3680,8 +3681,8 @@ impl PmixClient {
         let mut proc_guard = self
             .inner
             .proc
-            .lock()
-            .expect("pmix: client proc mutex poisoned");
+            .write()
+            .expect("pmix: client proc lock poisoned");
 
         match self.inner.state() {
             PmixClientState::Uninitialized => {}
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3515,7 +3516,7 @@ pub fn info_with_string_key(key: &str, value: &str) -> Info {
 // ─────────────────────────────────────────────────────────────────────────────
 
 use std::sync::atomic::{AtomicU8, Ordering};
-use std::sync::{Arc, Mutex, OnceLock};
+use std::sync::{Arc, Mutex, OnceLock, RwLock};
```
SedahsReview: Nit (non-blocking): with the proc field now an `RwLock`, `Mutex` is no longer used anywhere in lib.rs (the only remaining mention is a doc comment at line 23). The crate-level `#![allow(unused_imports)]` silently masks it — consider dropping `Mutex` from this import so the import list reflects the new read-mostly locking.

src/lib.rs
```diff
@@ -3680,8 +3681,8 @@ impl PmixClient {
         let mut proc_guard = self
             .inner
             .proc
-            .lock()
-            .expect("pmix: client proc mutex poisoned");
+            .write()
```
SedahsReview, line 3684: Nit (non-blocking): the doc comment on `connect` (line 3678) still reads "Serialized on the session mutex" — this is now an `RwLock` write lock (and was a `Mutex` before). Suggest rewording to "session lock" so the docs match the actual synchronization primitive. No functional impact.

SedahsReview (approved):

Reviewed commit 2039f079 against base 74f52cf (verified against the real base SHA, single-commit PR).

### Verdict
**Approve.** The Mutex→RwLock conversion is correct, complete, and behavior-preserving. No critical or warning-level issues; 2 minor suggestions inline.

### Behavior preservation
- All 5 `inner.proc` lock sites converted: connect/disconnect take `.write()`, proc/rank/proc_with_nspace take `.read()`; `require_proc()`/`require_rank()` delegate unchanged. Signatures, return types, and panic semantics are identical.
- The early `state() != Live → None` guard in `proc()`/`rank()` is safe: the field is `None` in every stable non-Live state (init = None, failed connect leaves None, disconnect clears to None before Dead), so the guard only tightens concurrent-race observations toward the documented "when live" contract. No test or documented pattern relies on the old race-dependent `Some`.
- `proc_with_nspace` deliberately keeps its `.read()` + `ok_or(ErrInit)` path — unchanged semantics (ErrInit when disconnected).
- No lock-order issue: `state` is a lock-free `AtomicU8`; the RwLock is the only lock, and readers never hold it while acquiring anything else.

### Test
`test_pmixclient_read_accessors_are_shared_across_threads` is deterministic (8 threads × 100 iters, immutable snapshot, rank always 0), compiles against the private-field access rules (test module is inside lib.rs), has the required `// SAFETY:` on `zeroed()`, and genuinely exercises concurrent shared readers. Passes.

### Quality gates (PMIX_INCLUDE_DIR=/tmp/openpmix-build/include, PMIX_LIB_DIR=/tmp/openpmix-build/src/.libs, OpenPMIx 6.1.0)
- cargo build: PASS
- cargo test --lib pmixclient -- --test-threads=1: 7/7 PASS (incl. new test)
- cargo test --lib -- --test-threads=1: 1832 passed, 0 failed, 29 ignored
- cargo clippy --lib -- -D warnings: 6 errors, all pre-existing `missing_safety_doc` in bindgen-generated bindings.rs; none in src/lib.rs changed lines
- cargo fmt --check: only pre-existing repo-wide diffs; PR-changed regions are fmt-clean
- cargo check --features mock_ffi --lib: PASS (44 warnings, baseline)

1. `src/lib.rs:3519` — `Mutex` is now unused in the `use std::sync::...` import (crate-level `#![allow(unused_imports)]` masks it); drop it.
2. `src/lib.rs:3678` doc comment still says "Serialized on the session mutex" — the proc lock is now an RwLock; consider "session lock" for accuracy.

SedahsReview:
## Code Review Summary — APPROVED (review id 4889735415)

Reviewed commit `2039f0798` (base `74f52cf3`, single-commit PR). Token: SedahsReview.

### Verdict
**APPROVE** — Mutex→RwLock conversion is correct, complete, and behavior-preserving. No critical or warning-level issues.

### Behavior preservation
- **All 5 `inner.proc` lock sites converted**: connect/disconnect use `.write()`; `proc()`/`rank()`/`proc_with_nspace()` use `.read()`; `require_proc()`/`require_rank()` delegate unchanged. Signatures, return types, and panic semantics identical.
- **Early Live-state check in `proc()`/`rank()` is not a regression**: traced the lifecycle invariant — the field is `None` in every stable non-Live state (init `None`; failed connect leaves `None` since it's only set on success; disconnect clears to `None` before `Dead`). The guard only changes concurrent-race observations (returns `None` instead of a race-dependent `Some`), which matches the documented "when live" contract. No test or documented pattern depends on the old race behavior.
- **`proc_with_nspace()` deliberately keeps `.read()` + `ok_or(ErrInit)`** — unchanged ErrInit-when-disconnected semantics.
- **No lock-order concern**: `state` is a lock-free `AtomicU8`; the RwLock is the only lock and readers never hold it while acquiring anything else. Readers block briefly behind connect/disconnect writes, same as before.
- **Poison messages** consistently renamed to "pmix: client proc lock poisoned"; no stale "proc mutex poisoned" strings for this lock.

### Critical
None.

### Warnings
None.

### Suggestions (non-blocking, posted inline)
- `src/lib.rs:3519` — `Mutex` now unused in the import (masked by crate-level `#![allow(unused_imports)]`); drop it.
- `src/lib.rs:3678` — doc comment still says "Serialized on the session mutex"; suggest "session lock" for accuracy.

### Looks Good
- Deterministic concurrent test (8 threads × 100 iters, immutable snapshot, no timing deps), with the required `// SAFETY:` on `zeroed()`.
- Crate-level doc and Debug string updated; `Send + Sync` preserved (existing `test_pmixclient_send_sync_clone` passes).

### Quality gates (OpenPMIx 6.1.0 via PMIX_INCLUDE_DIR/PMIX_LIB_DIR)
- `cargo build`: PASS
- `cargo test --lib pmixclient -- --test-threads=1`: 7/7 PASS (incl. new threading test)
- `cargo test --lib -- --test-threads=1`: **1832 passed, 0 failed, 29 ignored**
- `cargo clippy --lib -- -D warnings`: 6 errors, all pre-existing `missing_safety_doc` in bindgen-generated `bindings.rs`; **none in src/lib.rs changed lines**
- `cargo fmt --check`: only pre-existing repo-wide diffs; PR-changed regions fmt-clean
- `cargo check --features mock_ffi --lib`: PASS (44 warnings, baseline)

---
*Reviewed by Hermes Agent (code-reviewer profile)*

PR #398, opened 2026-08-08
perf: fold _nb SEQ/REGISTRY pairs into atomic-backed Registry structs

### LINKED ISSUE #386: perf: _nb callback registries serialize every non-blocking op on 2-3 process-global Mutexes (SEQ counter + registry map)
## Summary

Every non-blocking API in this crate pays **2–3 global `Mutex` acquisitions per call**, and because the locks are process-global, **all threads serialize** on them. This directly undercuts the crate's multi-threaded selling point (OpenPMIx ≥ 6.1 threadshift is the point of the `PmixClient: Clone + Send + Sync` design).

## Evidence

Per `_nb` op the pattern is (e.g. `src/data_ops/mod.rs`):

1. `GET_SEQ.lock()` → `*seq += 1` → unlock (`src/data_ops/mod.rs:310`)
2. `GET_REGISTRY.lock()` → `insert` → unlock (`:315`)
3. FFI call; on immediate failure a **third** lock of `GET_REGISTRY` to `remove` (`:378`)
4. Completion bridge: `GET_REGISTRY.lock()` → `remove` → unlock (`:246`)

~35 static `SEQ`/`REGISTRY` pairs across `data_ops` (5), `groups` (5), `query_log` (2), `security` (3), `allocation` (3), `monitoring` (1), `process_mgmt` (3), `server/mod.rs` (10), `server/data.rs` (3), `server/pset.rs` (2), `utility` IOF (1).

## Impact

- A 128-thread client issuing concurrent `get_nb`/`lookup`/`publish`/`fence_nb` ops contends on one global `Mutex<usize>` (the SEQ counter) plus one global registry `Mutex` for every op. The SEQ lock is pure overhead — it protects a counter that is trivially atomic.
- `monitoring` even splits seq into a separate `Mutex<u64>` (`src/monitoring.rs:134`).

## Correctness sub-issue (same refactor)

- `data_ops`, `groups`, `query_log`, `security`, `allocation`, `server/*` increment via raw `*seq += 1` (~30 sites); a wrap to `0` after `usize::MAX` requests would produce a **null cbdata** (`cbdata::encode_req_id` debug-asserts non-zero). `cbdata::next_req_id` (saturating, `src/cbdata.rs:56`) exists but is only used by 4 modules (`groups`, `process_mgmt`, `server/data`, `cbdata`).
- `src/data_ops/mod.rs:1356` and `:320` still carry the stale "shift left by 2" comment from the pre-`cbdata::encode_req_id` era.

## Suggested fix

- Replace every `Mutex<usize>`/`Mutex<u64>` SEQ with `AtomicU64::fetch_add(1, Relaxed)` (wrapping is acceptable with a `debug_assert` on non-zero, or keep saturating via a small helper).
- Fold SEQ + registry map into a single `struct Registry { seq: AtomicU64, map: Mutex<HashMap<u64, …>> }` so an op takes **one** lock instead of two, and the immediate-failure path can reuse the same guard.
- Route all modules through `cbdata::next_req_id` semantics.

Related: `THREADING.md` §8 defers "Global FFI mutex (#53)" — this issue is about the *Rust-side* registries, which are independent of that decision.

SedahsDev:
Closes #386

## Summary

Replaces the crate's ~35 process-global `Mutex<usize>`/`Mutex<u64>` SEQ counters + `Mutex<HashMap>` registry pairs with a single atomic-backed `Registry<T>` (new in `src/cbdata.rs`):

- **Lock-free request-ID sequence**: `AtomicU64` + `compare_exchange_weak` (saturating, never 0), so allocating a request ID never takes the callback-map lock. An `_nb` op now pays ONE map lock acquisition instead of two (SEQ + map), and concurrent threads never serialize on the counter.
- **One-lock-per-op**: `Registry::insert_next(cb)` allocates the ID and inserts the callback under a single map lock; the failure path reuses the same registry (no third lock). Completion bridges are unchanged (`Registry::lock()` returns the same `MutexGuard<HashMap>` shape).
- **Correctness fix**: every former raw `*seq += 1` site (which could wrap to 0 → null cbdata) now goes through the saturating atomic path; `monitoring`'s `Mutex<u64>` split is unified with the common `usize` approach.
- **Stale comments removed**: the pre-`cbdata::encode_req_id` era "shift left by 2" comments in `src/data_ops/mod.rs` are gone.
- **Legacy helpers deleted**: `cbdata::next_req_id(&Mutex<usize>)`, `encode_req_id_u64`, `decode_req_id_u64` (and their tests) once the last callers were converted.

## Modules converted

`data_ops` (5 pairs), `groups` (5), `process_mgmt` (3), `query_log` (2), `security` (3), `allocation` (3), `monitoring` (1), `server/data.rs` (3), `server/pset.rs` (2), `server/mod.rs` (10). Utility IOF's registry has no sequence and was left unchanged.

## Test plan

- Full local lib suite (mock FFI, no daemon): 1820 passed, 0 failed.
- New `cbdata::Registry` unit tests: ID sequence starts at 1 and never returns 0; `insert_next`/`remove` round-trip; 8-thread concurrent uniqueness (800 unique IDs).
- All existing `_nb` tests across converted modules pass unchanged; obsolete `*_seq_increments` tests (which asserted the removed Mutex counters) were deleted.
- Daemon/DVM tests remain CI-only.

### DIFF
--- a/src/cbdata.rs
+++ b/src/cbdata.rs
@@ -12,7 +12,7 @@
 
 use std::collections::HashMap;
 use std::os::raw::c_void;
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Mutex, MutexGuard};
 
 /// Encode a non-zero request ID as opaque cbdata for a PMIx C callback.
@@ -37,7 +37,7 @@ pub fn decode_req_id(cbdata: *mut c_void) -> usize {
 
 /// Process-global callback registry with a lock-free request-ID sequence.
 ///
-/// The sequence counter is an `AtomicU64`, so allocating a request ID never
+/// The sequence counter is an `AtomicUsize`, so allocating a request ID never
 /// takes the callback-map lock; concurrent ops therefore never serialize on
 /// the counter. An op pays ONE map lock acquisition instead of the previous
 /// SEQ lock + map lock pair.
@@ -45,7 +45,7 @@ pub fn decode_req_id(cbdata: *mut c_void) -> usize {
 /// Request IDs start at 1 and saturate rather than wrap, so they can never
 /// become 0 (which would encode as a null cbdata pointer).
 pub struct Registry<T> {
-    seq: AtomicU64,
+    seq: AtomicUsize,
     map: Mutex<HashMap<usize, T>>,
 }
 
@@ -59,7 +59,7 @@ impl<T> Registry<T> {
     /// Create an empty registry with the sequence starting at zero.
     pub fn new() -> Self {
         Self {
-            seq: AtomicU64::new(0),
+            seq: AtomicUsize::new(0),
             map: Mutex::new(HashMap::new()),
         }
     }
@@ -75,7 +75,7 @@ impl<T> Registry<T> {
                 .compare_exchange_weak(current, next, Ordering::Relaxed, Ordering::Relaxed)
                 .is_ok()
             {
-                return next as usize;
+                return next;
             }
         }
     }
--- a/src/cbdata.rs
+++ b/src/cbdata.rs
@@ -10,8 +10,10 @@
 //! 2. Use [`std::ptr::with_exposed_provenance_mut`] / [`pointer::addr`] so the
 //!    conversion is explicit and portable.
 
+use std::collections::HashMap;
 use std::os::raw::c_void;
-use std::sync::Mutex;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::{Mutex, MutexGuard};
 
 /// Encode a non-zero request ID as opaque cbdata for a PMIx C callback.
 ///
@@ -59,6 +61,72 @@ pub fn next_req_id(seq: &Mutex<usize>) -> usize {
     *g
 }
 
+/// Process-global callback registry with a lock-free request-ID sequence.
+///
+/// The sequence counter is an `AtomicU64`, so allocating a request ID never
+/// takes the callback-map lock; concurrent ops therefore never serialize on
+/// the counter. An op pays ONE map lock acquisition instead of the previous
+/// SEQ lock + map lock pair.
+///
+/// Request IDs start at 1 and saturate rather than wrap, so they can never
+/// become 0 (which would encode as a null cbdata pointer).
+pub struct Registry<T> {
+    seq: AtomicU64,
+    map: Mutex<HashMap<usize, T>>,
+}
+
+impl<T> Default for Registry<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<T> Registry<T> {
+    /// Create an empty registry with the sequence starting at zero.
+    pub fn new() -> Self {
+        Self {
+            seq: AtomicU64::new(0),
+            map: Mutex::new(HashMap::new()),
+        }
+    }
+
+    /// Allocate the next non-zero request ID (saturating, never 0).
+    #[inline]
+    pub fn next_req_id(&self) -> usize {
+        loop {
+            let current = self.seq.load(Ordering::Relaxed);
+            let next = current.saturating_add(1).max(1);
+            if self
+                .seq
+                .compare_exchange_weak(current, next, Ordering::Relaxed, Ordering::Relaxed)
+                .is_ok()
+            {
+                return next as usize;
+            }
+        }
+    }
+
+    /// Allocate a request ID and insert its callback under one map lock.
+    #[inline]
+    pub fn insert_next(&self, value: T) -> usize {
+        let req_id = self.next_req_id();
+        self.lock().insert(req_id, value);
+        req_id
+    }
+
+    /// Lock the callback map.
+    #[inline]
+    pub fn lock(&self) -> MutexGuard<'_, HashMap<usize, T>> {
+        self.map.lock().unwrap_or_else(|e| e.into_inner())
+    }
+
+    /// Remove a callback from the registry.
+    #[inline]
+    pub fn remove(&self, req_id: usize) -> Option<T> {
+        self.lock().remove(&req_id)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -83,5 +151,46 @@ mod tests {
         assert_eq!(next_req_id(&seq), 1);
         assert_eq!(next_req_id(&seq), 2);
         assert_eq!(next_req_id(&seq), 3);
+    }
+
+    #[test]
+    fn registry_req_ids_start_at_one_and_never_zero() {
+        let reg = Registry::<()>::new();
+        assert_eq!(reg.next_req_id(), 1);
+        assert_eq!(reg.next_req_id(), 2);
+        assert_eq!(reg.next_req_id(), 3);
+    }
+
+    #[test]
+    fn registry_insert_next_and_remove_roundtrip() {
+        let reg = Registry::new();
+        let id = reg.insert_next("callback");
+        assert_eq!(reg.remove(id), Some("callback"));
+        assert_eq!(reg.remove(id), None);
+    }
+
+    #[test]
+    fn registry_concurrent_req_ids_are_unique() {
+        use std::sync::Arc;
+        let reg = Arc::new(Registry::<()>::new());
+        let mut handles = Vec::new();
+        for _ in 0..8 {
+            let reg = Arc::clone(&reg);
+            handles.push(std::thread::spawn(move || {
+                let mut ids = Vec::new();
+                for _ in 0..100 {
+                    ids.push(reg.next_req_id());
+                }
+                ids
+            }));
+        }
+        let mut all: Vec<usize> = Vec::new();
+        for h in handles {
+            all.extend(h.join().unwrap());
+        }
+        all.sort_unstable();
+        all.dedup();
+        assert_eq!(all.len(), 800);
+        assert!(all.iter().all(|&id| id != 0));
     }
 }
--- a/src/cbdata.rs
+++ b/src/cbdata.rs
@@ -35,17 +35,6 @@ pub fn decode_req_id(cbdata: *mut c_void) -> usize {
 }
 
 
-/// Allocate the next non-zero request ID from a per-registry sequence counter.
-///
-/// Starts at 1 (never returns 0 / null cbdata). Saturates rather than wrapping
-/// so a pathological overflow cannot produce a null id.
-#[inline]
-pub fn next_req_id(seq: &Mutex<usize>) -> usize {
-    let mut g = seq.lock().unwrap_or_else(|e| e.into_inner());
-    *g = g.saturating_add(1).max(1);
-    *g
-}
-
 /// Process-global callback registry with a lock-free request-ID sequence.
 ///
 /// The sequence counter is an `AtomicU64`, so allocating a request ID never
@@ -128,14 +117,6 @@ mod tests {
     #[test]
     fn never_null_for_nonzero() {
         assert!(!encode_req_id(1).is_null());
-    }
-
-    #[test]
-    fn next_req_id_starts_at_one_and_increments() {
-        let seq = Mutex::new(0);
-        assert_eq!(next_req_id(&seq), 1);
-        assert_eq!(next_req_id(&seq), 2);
-        assert_eq!(next_req_id(&seq), 3);
     }
 
     #[test]
--- a/src/cbdata.rs
+++ b/src/cbdata.rs
@@ -34,21 +34,6 @@ pub fn decode_req_id(cbdata: *mut c_void) -> usize {
     cbdata.addr()
 }
 
-/// Encode a `u64` request ID (used by some monitoring paths).
-#[inline]
-pub fn encode_req_id_u64(req_id: u64) -> *mut c_void {
-    debug_assert!(
-        req_id != 0,
-        "request id must be non-zero to avoid null cbdata"
-    );
-    std::ptr::with_exposed_provenance_mut::<c_void>(req_id as usize)
-}
-
-/// Decode opaque cbdata into a `u64` request ID.
-#[inline]
-pub fn decode_req_id_u64(cbdata: *mut c_void) -> u64 {
-    cbdata.addr() as u64
-}
 
 /// Allocate the next non-zero request ID from a per-registry sequence counter.
 ///
--- a/src/cbdata.rs
+++ b/src/cbdata.rs
@@ -34,7 +34,6 @@ pub fn decode_req_id(cbdata: *mut c_void) -> usize {
     cbdata.addr()
 }
 
-
 /// Process-global callback registry with a lock-free request-ID sequence.
 ///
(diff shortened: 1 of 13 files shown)


src/query_log.rs
```diff
@@ -1086,16 +1061,13 @@ mod tests {
         // If Ok, the callback was NOT removed from registry (no PMIx server to
         // invoke the bridge callback). Clean it up ourselves.
         if result.is_ok() {
-            let req_id = {
-                let seq = QUERY_SEQ.lock().unwrap();
-                *seq
-            };
-            let mut registry = QUERY_REGISTRY.lock().unwrap();
+            let req_id = QUERY_REGISTRY.next_req_id();
```
SedahsReview: WARNING — off-by-one regression introduced by this adaptation. `query_info_nb` inserts the callback under the ID returned by `insert_next` (call it N). `next_req_id()` here returns **N+1** (and permanently advances the counter), so `registry.remove(&req_id)` is a no-op and the entry at N leaks. The old code read the seq value *without* incrementing (i.e. N), so removal was correct before this PR. If this test is ever run with a real PMIx server and the call returns Ok, the `assert_eq!(size_after, size_before)` at ~line 1073 will spuriously fail. The test is `#[ignore]`d so the suite is unaffected (1820/1820 green), but the cleanup logic is now wrong. Suggest capturing the actual inserted ID (track it at insert time, or scan for the entry whose callback matches) instead of allocating a fresh one.

src/cbdata.rs
```diff
@@ -32,31 +34,71 @@ pub fn decode_req_id(cbdata: *mut c_void) -> usize {
...
+    pub fn next_req_id(&self) -> usize {
+        loop {
+            let current = self.seq.load(Ordering::Relaxed);
+            let next = current.saturating_add(1).max(1);
+            if self
+                .seq
+                .compare_exchange_weak(current, next, Ordering::Relaxed, Ordering::Relaxed)
+                .is_ok()
+            {
+                return next as usize;
```
SedahsReview: SUGGESTION (non-blocking): `next as usize` truncates on 32-bit targets (u64 -> usize). The old `Mutex<usize>` counter saturated at usize::MAX and could never produce 0; here, once the counter passes 2^32 on a 32-bit build, the cast can yield 0 -> null cbdata, reintroducing the wrap-to-zero this PR is fixing. Harmless on 64-bit (the crate's realistic target). An `AtomicUsize` would match the map key type and the old semantics exactly while keeping the CAS loop unchanged.

src/server/mod.rs
```diff
@@ -1579,13 +1549,9 @@ pub fn server_dmodex_request(
     callback: Box<dyn DmodexRequestCallback>,
 ) -> Result<(), PmixStatus> {
     // Allocate a unique request ID and register the callback.
-    let req_id = {
-        let mut seq = DMODEX_REQUEST_SEQ.lock().unwrap();
-        *seq += 1;
-        *seq
-    };
+    let req_id = DMODEX_REQUEST_REGISTRY.next_req_id();
```
SedahsReview, line 1552: SUGGESTION (non-blocking): the PR removed the stale 'shift left by 2' comments in data_ops, but 8 sites of the same stale narrative remain on code this PR rewrote: plain comments at server/mod.rs:1559, 1825, 2008, 2220, and SAFETY comments at server/mod.rs:2470, 2669 and allocation.rs:384, 749. `encode_req_id` does not shift — it casts the request ID directly via `with_exposed_provenance_mut`; the never-null guarantee now comes from `Registry` never returning 0. The SAFETY comments misdescribe the actual mechanism and should be corrected or dropped.

SedahsReview (changes requested):
## Review verdict: REQUEST_CHANGES (1 warning; production code verified sound)

This is a clean, well-executed cross-module refactor. The atomic `Registry<T>` core is correct, all 10 module conversions preserve semantics, the full lib suite passes (1820/1820, 0 failed), and build/clippy/fmt are clean for the PR's files. One blocking warning: a logic regression introduced in an adapted (ignored) test's cleanup path in src/query_log.rs. Suggestions below are non-blocking.

### What was verified
- **CAS sequence is sound.** `next_req_id()`: Relaxed ordering is sufficient — the only invariant is distinct non-zero IDs; CAS linearizability guarantees at-most-one winner per value, `.max(1)` prevents 0, saturating prevents wrap. No ABA: the counter is monotonic (CAS from the observed `current`, never revisits a value). The only theoretical edge is saturation-collision at u64::MAX (2^64-1 allocations) and u64->usize truncation on 32-bit targets — both pre-existing-equivalent or theoretical; see inline suggestion.
- **insert_next ordering is correct**: the completion bridge only fires on a registered ID, and the FFI call happens after insertion, so there is no window where a callback fires before its registry entry exists.
- **All 10 modules converted consistently**: data_ops (5), groups (5), query_log (2), security (3), allocation (3), monitoring (1), process_mgmt (3), server/mod.rs (10), server/data.rs (3), server/pset.rs (2). Every old map VALUE type is preserved exactly (Box<dyn Trait>, callback wrappers); every `_SEQ` static is gone (grep: 0 remaining); no raw `*seq += 1` sites remain (grep: 0); failure paths and bridges keep the same registry-miss -> callback-not-invoked and remove-before-invoke semantics; poison handling folds into `Registry::lock()` via into_inner (matches the pattern process_mgmt already used). `PmixServerModule` is untouched. No public API signature changes.
- **monitoring**: the Mutex<u64> split is unified to the common usize approach; bridge now uses encode_req_id/decode_req_id; zero `as u64` casts remain.
- **Deleted tests**: all deleted tests were dead Mutex-counter assertions (`*_seq_increments`/`*_seq_is_accessible`); data_ops' were adapted to `next_req_id()` monotonicity instead. No meaningful coverage lost. New cbdata Registry tests (start-at-1/never-0, insert/remove round-trip, 8-thread x 100 uniqueness) are good.
- **Quality gates**: cargo build OK (OpenPMIx 6.1.0); cargo test --lib -- --test-threads=1 -> 1820 passed, 0 failed, 29 ignored; cargo clippy --lib -- -D warnings -> only the 6 pre-existing bindgen `missing_safety_doc` errors (all in generated out/bindings.rs, none in PR files); cargo fmt --check -> only pre-existing examples/ diffs, none in PR files.

### Warning
- src/query_log.rs:1064 — off-by-one in the ignored test's cleanup (see inline).

- src/cbdata.rs:78 — `next as usize` truncates on 32-bit targets; `AtomicUsize` would match the key type and old semantics exactly.
- src/server/mod.rs:1552 (+ 7 more sites) — stale 'shift left by 2' comments (incl. 4 SAFETY comments in server/mod.rs:2470/2669, allocation.rs:384/749) survive on code this PR rewrote; `encode_req_id` doesn't shift.
- Nit: allocation.rs / server/mod.rs call sites use next_req_id() + lock().insert() where insert_next() is the one-lock idiom used elsewhere (functionally identical).

SedahsReview:
## Code Review Summary — PR #398 (perf: atomic-backed Registry structs)

**Verdict: REQUEST_CHANGES** — 1 warning, 3 suggestions. The refactor itself is verified sound; the single blocker is a small logic regression in an adapted (ignored) test. Formal review submitted with inline comments.

### 🔴 Warning (blocking)
- **src/query_log.rs:1064** — off-by-one in `test_query_info_nb_callback_registration_and_cleanup` (`#[ignore]`d): `query_info_nb` inserts at ID N; the cleanup calls `next_req_id()` which returns **N+1**, so `remove(N+1)` is a no-op and the entry leaks; the size assert at ~:1073 would fail if the test ever runs with a real server and gets Ok. Old code read the seq value (N) without incrementing — this is a regression introduced by the adaptation. Fix is trivial; suite is unaffected (test is ignored).

### ⚠️ Warnings / issues checked and cleared
- CAS sequence: sound — Relaxed sufficient, no ABA (monotonic), `.max(1)`/saturating prevent 0/wrap. Only theoretical edges: saturation-collision at u64::MAX (pre-existing behavior) and u64→usize truncation on 32-bit (suggestion below).
- All 10 modules converted, value types preserved exactly, registry-miss → callback-not-invoked and remove-before-invoke semantics unchanged, `_SEQ` statics and raw `*seq += 1` sites all gone (grep-verified 0), no public API signature changes, `PmixServerModule` untouched, monitoring u64 split unified with no `as u64` casts left.

### 💡 Suggestions (non-blocking, see inline)
- **src/cbdata.rs:78** — `next as usize` truncates on 32-bit targets; `AtomicUsize` would match the map key type and old semantics exactly.
- **src/server/mod.rs:1552 + 7 more sites** — stale 'shift left by 2' comments (incl. 4 SAFETY comments at server/mod.rs:2470/2669, allocation.rs:384/749) survive on code this PR rewrote; `encode_req_id` doesn't shift.
- Nit: allocation.rs / server/mod.rs use `next_req_id()` + `lock().insert()` where `insert_next()` is the one-lock idiom used elsewhere.

### ✅ Looks Good
- `Registry<T>` API (new/Default/next_req_id/insert_next/lock/remove) is minimal and correct; `insert_next` has no callback-before-insert race.
- Deleted tests were dead Mutex-counter assertions only; data_ops' adapted to `next_req_id()` monotonicity; new cbdata Registry tests (start-at-1/never-0, round-trip, 8×100 concurrent uniqueness) are solid.
- Poison handling folded into `Registry::lock()` (into_inner) — matches the crate's prevailing pattern.

### Quality gates (local, PMIx 6.1.0 via /tmp/openpmix-build)
- `cargo build` — OK
- `cargo test --lib -- --test-threads=1` — **1820 passed, 0 failed, 29 ignored** (acceptance criterion met)
- `cargo clippy --lib -- -D warnings` — only the 6 pre-existing bindgen `missing_safety_doc` errors (generated out/bindings.rs; none in PR files)
- `cargo fmt --check` — only pre-existing examples/ diffs; none in PR files
- CI (GitHub Actions): Regular tests, Clippy, DVM tests all green; Daemon tests in progress at review time.

---

SedahsReview (approved):
## Review verdict: APPROVE — all 3 prior findings resolved, no new issues

Re-review of commit d2a251c (fixes for review 4889936695 / comments 3741806682-84). Every prior point checked against the current code and confirmed fixed:

**1. WARNING — query_log.rs off-by-one (3741806682): FIXED.** The ignored test `test_query_info_nb_callback_registration_and_cleanup` no longer calls `next_req_id()` (which would advance the sequence and return N+1). It now snapshots `ids_before: HashSet<usize>` of the registry keys before the call, and on `Ok` removes the exact key that is *not* in `ids_before` — i.e. precisely the ID `query_info_nb` inserted via `insert_next`. `size_before` is derived from the same snapshot, so the size assertion is sound if the test is ever un-ignored (under `--test-threads=1` nothing else mutates the registry mid-test). No off-by-one remains.

**2. SUGGESTION — u64→usize truncation (3741806683): FIXED.** `Registry<T>.seq` is now `AtomicUsize` (cbdata.rs:48), `next_req_id` computes `current.saturating_add(1).max(1)` and returns `next` from the successful `compare_exchange_weak` directly — no `as usize` cast, so no 32-bit truncation and the value can never be 0 (saturates at `usize::MAX` instead of wrapping). Map keys stay `usize`; `encode_req_id` casts via `with_exposed_provenance_mut` and `decode_req_id` via `ptr::addr`, consistent with the strict-provenance doc. All 5 cbdata unit tests still pass, including `registry_req_ids_start_at_one_and_never_zero` and the 8-thread uniqueness test.

**3. SUGGESTION — stale "shift left by 2" comments (3741806684): FIXED.** All 8 sites rewritten to the actual mechanism: plain comments at server/mod.rs:1559/1825/2008/2220 and SAFETY comments at server/mod.rs:2470/2669 + allocation.rs:384/749 now state that `encode_req_id` casts the request ID directly and that the Registry never returns 0, so cbdata is never null. Grep across all 13 PR files finds no remaining "shift left" / "alignable" narrative (the only "bit-shifting" mention is cbdata.rs's module doc, which *historically* describes the old approach — accurate and intentional).

**Cross-module check:** all 12 registry conversions (data_ops ×3, groups ×5, process_mgmt ×3, query_log ×2, security ×2, server ×4, monitoring, allocation ×2) follow the same pattern — `Registry<T>` replacing the SEQ mutex + `Mutex<HashMap>` pair, `insert_next`/`next_req_id` preserving the old first-ID-at-1 semantics, error paths still removing the registered callback. monitoring.rs's u64-keyed variant was correctly unified onto the usize Registry with no `_u64` encode/decode remnants anywhere.

**Quality gates (local, PMIX_PREFIX=~/.local/openpmix-6.1.0):**
- `cargo build`: clean.
- `cargo test --lib -- --test-threads=1`: **1820 passed, 0 failed, 29 ignored** (matches the orchestrator's head verification exactly).
- `cargo clippy --lib -- -D warnings`: 6 errors, all pre-existing bindgen `missing_safety_doc` in generated `target/.../out/bindings.rs` — none in PR files.
- `cargo fmt --check`: repo-wide baseline is already non-clean; the PR's head-only fmt diffs are confined to its 13 changed files and the total diff-site count dropped vs base (485 vs 507) — no new violations outside the PR's own files.

Non-blocking nit: a double blank line at cbdata.rs:36-37 (fmt) introduced by the rewrite. Trivial; can fold into a later fmt pass.

No new critical or warning-level issues. Approving.

SedahsReview:
## Code Review Summary (round 2) — re-review of d2a251c

**Verdict: APPROVED** (formal review 4889950544). All 3 items from the previous CHANGES_REQUESTED round are resolved.

### Critical
- None.

### Warnings
- None. The prior warning (query_log.rs:1064 off-by-one in the ignored `query_info_nb` cleanup test) is fixed: the test snapshots `ids_before` and removes the exact inserted ID instead of advancing the sequence; the size assertion is now sound.

### Suggestions (all previously raised, all fixed)
- cbdata.rs:78 — `Registry.seq` is now `AtomicUsize`; `next_req_id` returns the CAS result directly (no `as usize` cast, no 32-bit truncation, never 0 via `saturating_add(1).max(1)`).
- server/mod.rs + allocation.rs — all 8 stale "shift left by 2" comments rewritten to describe the actual `encode_req_id` direct-cast mechanism; no stale narrative remains in any PR file.

### New Issues
- None critical/warning. Non-blocking nit only: double blank line at cbdata.rs:36-37 (fmt).

### Looks Good
- Uniform `Registry<T>` conversion across all 12 modules (data_ops ×3, groups ×5, process_mgmt ×3, query_log ×2, security ×2, server ×4, monitoring, allocation ×2); monitoring's u64 variant correctly unified; error paths still remove registered callbacks; first-ID-at-1 semantics preserved.
- 5 cbdata unit tests cover roundtrip, non-null, start-at-1, insert/remove, and 8-thread uniqueness.

### Quality Gates (local, PMIX_PREFIX=~/.local/openpmix-6.1.0)
- `cargo build`: clean.
- `cargo test --lib -- --test-threads=1`: **1820 passed, 0 failed, 29 ignored**.
- `cargo clippy --lib -- -D warnings`: 6 pre-existing bindgen `missing_safety_doc` errors in generated bindings.rs only — none in PR files.
- `cargo fmt --check`: PR-attributable diffs confined to its 13 files; total diff-site count dropped vs base (485 vs 507).

Closes #386. Thanks for the clean follow-up commit.

---

PR #409, opened 2026-08-09
fix(data): return ErrBadParam from get_value on malformed key instead of panicking

### LINKED ISSUE #403: get_value panics on malformed key slice (CStr::from_bytes_with_nul().unwrap()) — should return Err(PmixError::ErrBadParam)
## Summary

`get_value(proc: &Proc, key: &[u8], info: Option<Info>) -> Result<PmixOwnedValue, PmixError>` (src/lib.rs:3825) panics on a malformed `key` slice instead of returning an error:

```rust
// src/lib.rs:3845
let key_ptr = CStr::from_bytes_with_nul(key).unwrap().as_ptr();
```

`CStr::from_bytes_with_nul` fails — and the `.unwrap()` panics — when:
- `key` is empty (no trailing NUL), or
- `key` contains an interior NUL byte before the terminator.

The function already returns `Result` and already maps other invalid input to `PmixError::ErrBadParam` (e.g. `proc_with_nspace` at :3812), so a malformed key should be `Err(PmixError::ErrBadParam)`.

(Line refs against `main` @ 81064c8. Note: the separate allocation-leak bug in `get_value` is tracked in #389; this issue is strictly about the panic path.)

## Suggested fix

```rust
let key_ptr = match CStr::from_bytes_with_nul(key) {
    Ok(c) => c.as_ptr(),
    Err(_) => return Err(PmixError::ErrBadParam),
};
```

## Repro (today)

```rust
let proc = Proc::new("ns", 0).unwrap();
get_value(&proc, b"pmix.host", None);     // fine (trailing NUL present)
get_value(&proc, b"", None);              // panics: called `Result::unwrap()` on an `Err` value
get_value(&proc, b"bad\0key\0", None);    // panics
```

## Impact

A public `Result`-returning data-access API has an unrecoverable panic path on caller-supplied bytes that it could cheaply reject with the error type it already uses.

SedahsDev:
Closes #403

## Summary
get_value now returns Err(PmixError::ErrBadParam) when the key slice is empty or contains an interior NUL byte, instead of panicking on CStr::from_bytes_with_nul().unwrap().

## Change
Replaced the unwrap on CStr::from_bytes_with_nul with explicit error handling that returns Err(PmixError::ErrBadParam). No FFI dispatch or unrelated get_value behavior was changed.

## Tests
Added unit coverage for empty keys, interior NUL keys, and a valid key through MockGuard/mock storage.

## Test plan
- Local: PMIX_PREFIX=$HOME/.local/openpmix-6.1.0 cargo build
- Local: PMIX_PREFIX=$HOME/.local/openpmix-6.1.0 cargo test --lib get_value
- Local: git diff --check
- Daemon-based tests are left to CI because no PMIx daemon/prterun is available locally.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4154,8 +4154,10 @@ mod tests {
     fn test_get_value_accepts_valid_key_with_mock() {
         let _guard = crate::mock_ffi::MockGuard::new();
         let proc = Proc::new("ns", 0).unwrap();
-        crate::mock_ffi::mock_store_value("pmix.host", b"host", PMIX_STRING as u32);
-        assert!(get_value(&proc, b"pmix.host\0", None).is_ok());
+        crate::mock_ffi::mock_store_value("pmix.host", b"host", PMIX_STRING);
+        let result = get_value(&proc, b"pmix.host\0", None);
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap().uint32(), 1);
     }
 
     // ──────────────────────────────────────────────────────────────────────
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3866,7 +3866,10 @@ pub fn get_value(proc: &Proc, key: &[u8], info: Option<Info>) -> Result<PmixOwne
         }
     }
 
-    let key_ptr = CStr::from_bytes_with_nul(key).unwrap().as_ptr();
+    let key_ptr = match CStr::from_bytes_with_nul(key) {
+        Ok(c) => c.as_ptr(),
+        Err(_) => return Err(PmixError::ErrBadParam),
+    };
     let status = PmixStatus::from_raw(crate::pmix_ffi_or_mock!(
         mock = unsafe {
             crate::mock_ffi::mock_get(
@@ -4119,6 +4122,41 @@ mod tests {
     }
 
     use super::*;
+
+    // ──────────────────────────────────────────────────────────────────────
+    // get_value
+    // ──────────────────────────────────────────────────────────────────────
+
+    fn synthetic_proc() -> Proc {
+        Proc {
+            // SAFETY: get_value does not inspect the process fields before
+            // validating the key in these malformed-input tests.
+            handle: unsafe { std::mem::zeroed() },
+            len: 1,
+        }
+    }
+
+    #[test]
+    fn test_get_value_rejects_empty_key() {
+        let proc = synthetic_proc();
+        let result = get_value(&proc, b"", None);
+        assert!(matches!(result, Err(PmixError::ErrBadParam)));
+    }
+
+    #[test]
+    fn test_get_value_rejects_interior_nul_key() {
+        let proc = synthetic_proc();
+        let result = get_value(&proc, b"bad\0key\0", None);
+        assert!(matches!(result, Err(PmixError::ErrBadParam)));
+    }
+
+    #[test]
+    fn test_get_value_accepts_valid_key_with_mock() {
+        let _guard = crate::mock_ffi::MockGuard::new();
+        let proc = Proc::new("ns", 0).unwrap();
+        crate::mock_ffi::mock_store_value("pmix.host", b"host", PMIX_STRING as u32);
+        assert!(get_value(&proc, b"pmix.host\0", None).is_ok());
+    }
 
     // ──────────────────────────────────────────────────────────────────────
     // get_version


src/lib.rs
```diff
@@ -4120,6 +4123,41 @@ mod tests {
...
+        let result = get_value(&proc, b"bad\0key\0", None);
+        assert!(matches!(result, Err(PmixError::ErrBadParam)));
+    }
+
+    #[test]
+    fn test_get_value_accepts_valid_key_with_mock() {
+        let _guard = crate::mock_ffi::MockGuard::new();
+        let proc = Proc::new("ns", 0).unwrap();
+        crate::mock_ffi::mock_store_value("pmix.host", b"host", PMIX_STRING as u32);
+        assert!(get_value(&proc, b"pmix.host\0", None).is_ok());
```
SedahsReview: Suggestion (non-blocking): `test_get_value_accepts_valid_key_with_mock` asserts only `.is_ok()`. Since `mock_get`'s fallback path (key not found in the mock store) also returns `Ok` with a default value, consider also asserting the returned value's type/payload (e.g. `PMIX_STRING`) so the test proves the stored `pmix.host` entry is actually returned, not just that the mock call succeeded.

SedahsReview (approved):
## Verdict: APPROVE

Reviewed against head `868e1d3a` (base `7f3a771`). The fix is exactly the change described in the PR and issue #403: `get_value` now returns `Err(PmixError::ErrBadParam)` instead of panicking when the key slice is empty or contains an interior NUL.

**Correctness:** `CStr::from_bytes_with_nul` returns `Err` for both malformed-input classes (empty slice has no terminator; interior NUL precedes the terminator). The `return Err(...)` fires before any FFI dispatch, so no PMIx call is made with a bogus pointer. For valid keys the `Ok(c) => c.as_ptr()` arm has identical pointer semantics to the old `unwrap().as_ptr()`: the `&CStr` borrows `key`, `as_ptr` points into `key`'s buffer, and `key` (a `&[u8]` parameter) stays alive through the `PMIx_Get`/`mock_get` call — the temporary lifetime of the `CStr` reference is irrelevant because it owns nothing. Error type matches existing convention (`proc_with_nspace`, `multicluster_nspace_construct`).

**Scope:** Nothing else in `get_value` changed — the `pmix_ffi_or_mock!` dispatch, the `Box::from_raw` reclaim, and the status→error mapping are byte-identical to base. `put_value`/`fence`/`info_with_string_key` are untouched, and the separate allocation-leak issue #389 is correctly left out of scope.

**Tests:** The three new tests cover the exact panic paths from #403 (empty key, interior-NUL key) plus a valid key through the mock path. `synthetic_proc`'s zeroed handle is safe because the malformed-key paths return before `proc.handle` is used. The mock path is sound: `mock_get` returns a `Box::into_raw` allocation that `get_value`'s `Box::from_raw` reclaims exactly once.

Two non-blocking suggestions are attached inline / summarized in the PR comment; neither blocks merge. All CI check-runs on the head commit are green.

SedahsReview:
## Code Review Summary

**Verdict: APPROVE** — no critical or warning-level issues.

### Critical
None.

### Warnings
None.

### Suggestions
- **src/lib.rs:4158** (`test_get_value_accepts_valid_key_with_mock`) — asserts only `.is_ok()`. The mock's fallback path (key absent from store) also returns `Ok`, so asserting the returned value's type/payload (e.g. `PMIX_STRING`) would prove the stored `pmix.host` entry is actually returned. Non-blocking.
- **src/lib.rs:4157** — `PMIX_STRING as u32` cast is redundant; the constant is already `u32` (existing callers pass it uncast, e.g. `mock_ffi.rs` tests). Non-blocking.

### Looks Good
- Malformed keys (`b""`, `b"bad\0key\0"`) now return `Err(PmixError::ErrBadParam)` before any FFI call — the panic path from issue #403 is eliminated and the error type matches existing convention (`proc_with_nspace`, `multicluster_nspace_construct`).
- `Ok(c) => c.as_ptr()` is pointer-identical to the old `unwrap().as_ptr()` for valid keys; the `&CStr` borrows `key`, which stays alive through the FFI call, so lifetime semantics are unchanged.
- Scope is minimal: the `pmix_ffi_or_mock!` dispatch, `Box::from_raw` reclaim, and status→error mapping in `get_value` are untouched; `put_value`/`fence`/`info_with_string_key` untouched; issue #389 (allocation leak) correctly left out of scope.
- Tests are mock-based (no daemon), cover both malformed-input classes plus a valid-key happy path, and `synthetic_proc` is safe because the error paths return before `proc.handle` is dereferenced.

### Build / Test / Clippy / Fmt (local, PMIX_PREFIX=$HOME/.local/openpmix-6.1.0)
- `cargo build` — PASS.
- `cargo test --lib` — 1839 passed / 0 failed / 29 ignored. First run had one failure in `events::tests::test_pending_registration_delivery_and_rekey` (a timing-sensitive test in untouched `src/events.rs`); it passes in isolation and the full suite is green on re-run — pre-existing flakiness, not caused by this PR.
- `cargo clippy --lib -- -D warnings` — 6 errors, all in bindgen-generated `OUT_DIR/.../bindings.rs` (pre-existing environmental noise); none in `src/lib.rs`. CI Clippy check-run on the head commit is green.
- `cargo fmt --check` — 583 pre-existing diffs (present on main); verified via a `cargo fmt` apply-and-restore experiment that none of the PR's added lines are reformatted.
- `git diff --check origin/main...HEAD` — clean.
- GitHub CI on head `868e1d3a`: all 4 check-runs success (Regular tests, Daemon tests, DVM tests, Clippy).

Reviewed by SedahsReview (code-reviewer profile).

PR #455, opened 2026-08-12
fix(data_ops): zero pdata value source in nb lookup bridge to prevent double-free

### LINKED ISSUE #434: critical: nb lookup bridge double-frees pdata values
## Severity: critical (double-free)

`src/data_ops/mod.rs:774-793` (`lookup_callback_bridge`).

The bridge takes ownership of each `pdata.value` payload with `ptr::read(&pdata_ref.value)` but never zeroes the source, then calls `ffi::PMIx_Pdata_free(data, ndata)`. `PMIX_PDATA_FREE` destructs every element, running `PMIX_VALUE_DESTRUCT` which frees the payload — the same payload now owned by the `PmixOwnedValue` in results, whose `Drop` frees it again. The blocking `lookup()` (lines 661-663) explicitly zeroes the source and sets `PMIX_UNDEF` before destruct; the nb path omits this.

**Fix:** mirror the blocking path — after `ptr::read`, zero the source value and set `type_ = PMIX_UNDEF` before `PMIx_Pdata_free`.

SedahsDev:
Closes #434

## Summary
- Zero and mark each transferred non-blocking lookup pdata value as PMIX_UNDEF before PMIx cleanup.
- Add a mock-based regression test covering value transfer through the lookup callback bridge.

## Affected function
`lookup_callback_bridge` in `src/data_ops/mod.rs`.

## Rationale
The bridge moves value payload ownership into `PmixOwnedValue`; marking the source `PMIX_UNDEF` prevents `PMIx_Pdata_free` from destructing the same payload a second time.

## Test plan
- Local mock-based callback regression test (no PMIx daemon required).
- `cargo build` with `PMIX_PREFIX=$HOME/.local/openpmix-6.1.0`.
- `cargo test --lib` passes.
- Full `cargo test` and `cargo clippy --all-targets -- -D warnings` remain blocked by pre-existing repository test/lint failures; daemon tests require CI infrastructure.
- `cargo fmt --check` remains blocked by pre-existing formatting diffs outside this change.

### DIFF
--- a/src/data_ops/mod.rs
+++ b/src/data_ops/mod.rs
@@ -773,6 +773,8 @@ extern "C" fn lookup_callback_bridge(
                 let pmix_undef: ffi::pmix_data_type_t = ffi::PMIX_UNDEF as u16;
                 let value = if pdata_ref.value.type_ != pmix_undef {
                     let val = ptr::read(&pdata_ref.value);
+                    // SAFETY: `val` owns the payload copied from the PMIx result;
+                    // clear the source before PMIx frees the pdata array.
                     std::ptr::write_bytes(&mut pdata_ref.value, 0, 1);
                     pdata_ref.value.type_ = pmix_undef;
                     Some(PmixOwnedValue { inner: val, 
--- a/src/data_ops/mod.rs
+++ b/src/data_ops/mod.rs
@@ -756,7 +756,7 @@ extern "C" fn lookup_callback_bridge(
         unsafe {
             for i in 0..ndata {
                 let pdata = data.add(i);
-                let pdata_ref = &*pdata;
+                let pdata_ref = &mut *pdata;
 
                 let nspace_str = std::ffi::CStr::from_ptr(pdata_ref.proc_.nspace.as_ptr())
                     .to_string_lossy()
@@ -773,6 +773,8 @@ extern "C" fn lookup_callback_bridge(
                 let pmix_undef: ffi::pmix_data_type_t = ffi::PMIX_UNDEF as u16;
                 let value = if pdata_ref.value.type_ != pmix_undef {
                     let val = ptr::read(&pdata_ref.value);
+                    std::ptr::write_bytes(&mut pdata_ref.value, 0, 1);
+                    pdata_ref.value.type_ = pmix_undef;
                     Some(PmixOwnedValue { inner: val, 
             _not_thread_safe: std::marker::PhantomData,
         })
--- a/src/data_ops/tests.rs
+++ b/src/data_ops/tests.rs
@@ -2151,6 +2151,44 @@ use super::*;
 
     // ─── Mock-aware lookup_nb callback tests ────────────────────────────────
 
+    /// Verify lookup callback ownership survives PMIx freeing the pdata array.
+    #[test]
+    fn test_lookup_callback_bridge_transfers_value_before_pdata_free() {
+        use std::sync::atomic::{AtomicBool, Ordering};
+        static RECEIVED: AtomicBool = AtomicBool::new(false);
+
+        struct Callback;
+        impl LookupCallback for Callback {
+            fn on_result(self: Box<Self>, _status: PmixStatus, mut data: Vec<PmixPdata>) {
+                RECEIVED.store(
+                    data.pop().and_then(|pdata| pdata.value).is_some(),
+                    Ordering::SeqCst,
+                );
+            }
+        }
+
+        RECEIVED.store(false, Ordering::SeqCst);
+        let req_id = LOOKUP_REGISTRY.next_req_id();
+        LOOKUP_REGISTRY.lock().insert(req_id, Box::new(Callback));
+
+        let data = unsafe {
+            let data = libc::calloc(1, std::mem::size_of::<ffi::pmix_pdata_t>())
+                as *mut ffi::pmix_pdata_t;
+            assert!(!data.is_null());
+            (*data).value.type_ = PMIX_STRING_U16;
+            (*data).value.data.string = std::ffi::CString::new("value").unwrap().into_raw();
+            data
+        };
+
+        lookup_callback_bridge(
+            PMIX_SUCCESS,
+            data,
+            1,
+            crate::cbdata::encode_req_id(req_id),
+        );
+        assert!(RECEIVED.load(Ordering::SeqCst));
+    }
+
     /// Test lookup_nb callback bridge with success status.
     #[test]
     fn test_lookup_callback_bridge_success() {
--- a/src/data_ops/tests.rs
+++ b/src/data_ops/tests.rs
@@ -2175,7 +2175,7 @@ use super::*;
             let data = libc::calloc(1, std::mem::size_of::<ffi::pmix_pdata_t>())
                 as *mut ffi::pmix_pdata_t;
             assert!(!data.is_null());
-            (*data).value.type_ = PMIX_STRING as _;
+            (*data).value.type_ = crate::ffi::PMIX_STRING as _;
             let string = std::ffi::CString::new("value").unwrap();
             (*data).value.data.string = libc::strdup(string.as_ptr());
             assert!(!(*data).value.data.string.is_null());
--- a/src/data_ops/tests.rs
+++ b/src/data_ops/tests.rs
@@ -2175,8 +2175,10 @@ use super::*;
             let data = libc::calloc(1, std::mem::size_of::<ffi::pmix_pdata_t>())
                 as *mut ffi::pmix_pdata_t;
             assert!(!data.is_null());
-            (*data).value.type_ = PMIX_STRING_U16;
-            (*data).value.data.string = std::ffi::CString::new("value").unwrap().into_raw();
+            (*data).value.type_ = PMIX_STRING as _;
+            let string = std::ffi::CString::new("value").unwrap();
+            (*data).value.data.string = libc::strdup(string.as_ptr());
+            assert!(!(*data).value.data.string.is_null());
             data
         };
 


src/data_ops/tests.rs
```diff
@@ -2151,6 +2151,44 @@ use super::*;
...
+
+        RECEIVED.store(false, Ordering::SeqCst);
+        let req_id = LOOKUP_REGISTRY.next_req_id();
+        LOOKUP_REGISTRY.lock().insert(req_id, Box::new(Callback));
+
+        let data = unsafe {
+            let data = libc::calloc(1, std::mem::size_of::<ffi::pmix_pdata_t>())
+                as *mut ffi::pmix_pdata_t;
+            assert!(!data.is_null());
+            (*data).value.type_ = PMIX_STRING_U16;
```
SedahsReview: **Warning — test does not actually catch the double-free.** `PMIX_STRING_U16` is not freed by `PmixOwnedValue::drop` (its `free_value` falls through to the `_ => {}` arm in lib.rs:2957, which only handles STRING/PROC/BYTE_OBJECT/ENVAR/DATA_ARRAY). So even with the source-zeroing fix removed, this test passes: `PMIx_Pdata_free` frees the string once, and `PmixOwnedValue::drop` frees nothing, so there is no double-free to trip on. I verified this empirically by reverting the fix and re-running the test — it still passes. As written, the test also leaks the CString.

Use `PMIX_STRING` instead of `PMIX_STRING_U16` for the transferred value. `free_value` DOES free a `PMIX_STRING` payload, so: without the fix `PMIx_Pdata_free` frees the string and `PmixOwnedValue::drop` frees it again → glibc abort (test catches the regression); with the fix the payload is freed exactly once. This makes the regression test faithful to its stated purpose.

src/data_ops/mod.rs
```diff
@@ -773,6 +773,8 @@ extern "C" fn lookup_callback_bridge(
                 let pmix_undef: ffi::pmix_data_type_t = ffi::PMIX_UNDEF as u16;
                 let value = if pdata_ref.value.type_ != pmix_undef {
                     let val = ptr::read(&pdata_ref.value);
+                    std::ptr::write_bytes(&mut pdata_ref.value, 0, 1);
```
SedahsReview, line 778: **Suggestion (non-blocking):** the blocking `lookup()` path documents this copy-then-zero ownership transfer with a `// SAFETY:` comment (mod.rs:659-660). For consistency, add the same comment above these two lines explaining that ownership of the payload is moved into `PmixOwnedValue` and the source is zeroed so `PMIx_Pdata_free`'s destructor does not free it a second time.

SedahsReview (changes requested):
## Code Review — Request Changes

The core fix in `lookup_callback_bridge` (src/data_ops/mod.rs) is **correct**: after `ptr::read(&pdata_ref.value)`, the source is zeroed (`write_bytes`) and `type_` set to `PMIX_UNDEF` before `PMIx_Pdata_free`, exactly mirroring the blocking `lookup()` path (mod.rs:656-670). This correctly prevents the payload now owned by `PmixOwnedValue` from being freed a second time by the C destructor. No double-free, no leak in the fixed path. The blocking path is unchanged.

I verified build + tests locally (PMIX_PREFIX=$HOME/.local/openpmix-6.1.0): `cargo build` clean; `cargo test --lib` = 1849 passed / 0 failed. Clippy's 6 errors are all pre-existing bindgen `missing_safety_doc` in generated bindings.rs (unrelated). fmt diffs are all pre-existing (including in the untouched blocking path); the PR's added lines are fmt-clean.

One warning-level issue: the new regression test does not actually exercise the double-free it claims to guard against — see inline comment on tests.rs:2178. With the fix reverted, `test_lookup_callback_bridge_transfers_value_before_pdata_free` still passes (verified empirically), because `PmixOwnedValue::drop` → `free_value` does not free a `PMIX_STRING_U16` payload (falls through to the `_ => {}` arm in lib.rs:2957). So the test gives false confidence that the double-free is prevented, and it leaks the CString allocation. Using `PMIX_STRING` instead would make `free_value` free the payload on Drop, so reverting the fix would then produce a real double-free (glibc abort) that the test catches.

Also a minor suggestion (mod.rs:776): add a `// SAFETY:` comment mirroring the one on the blocking path (mod.rs:659-660) explaining the copy-then-zero ownership transfer.

SedahsReview:
## Code Review Summary

**Verdict: Changes Requested** (1 warning, 1 suggestion) — posted as formal review (id 4912547616).

### 🔴 Critical
- None.

### ⚠️ Warnings
- **src/data_ops/tests.rs:2178** — The new regression test does not actually catch the double-free it claims to prevent. It sets the transferred value's type to `PMIX_STRING_U16`, but `PmixOwnedValue::drop` → `free_value` does not free `PMIX_STRING_U16` payloads (falls through to the `_ => {}` arm in lib.rs:2957). I verified empirically: reverting the source-zeroing fix and re-running the test still gives `ok` (no double-free manifests, since the Rust side never frees that payload). The test also leaks the CString. Fix: use `PMIX_STRING` instead, which `free_value` DOES free — then without the fix `PMIx_Pdata_free` frees the string and `PmixOwnedValue::drop` frees it again → glibc abort (caught); with the fix the payload is freed exactly once.

### 💡 Suggestions
- **src/data_ops/mod.rs:776** — Add a `// SAFETY:` comment above the `write_bytes` + `type_ = PMIX_UNDEF` lines, mirroring the one on the blocking path (mod.rs:659-660), to document the copy-then-zero ownership transfer.

### ✅ Looks Good
- The fix itself is correct: after `ptr::read(&pdata_ref.value)` the source is zeroed and marked `PMIX_UNDEF` before `PMIx_Pdata_free`, exactly mirroring the blocking `lookup()` path (mod.rs:656-670). Prevents the double-free. No leak in the fixed path.
- Blocking `lookup()` path unchanged.
- No secrets, no debug leftovers, no merge-conflict markers.

### Build / Test / Lint
- `cargo build` (PMIX_PREFIX=$HOME/.local/openpmix-6.1.0): **pass**
- `cargo test --lib`: **1849 passed, 0 failed** (incl. the new test)
- `cargo clippy --lib -- -D warnings`: 6 errors, all pre-existing bindgen `missing_safety_doc` in generated `bindings.rs` — none from this PR
- `cargo fmt --check`: pre-existing repo-wide diffs only (incl. in the untouched blocking path); the PR's added lines are fmt-clean

---
*Reviewed by Hermes Agent (code-reviewer).*

src/data_ops/tests.rs
```diff
@@ -2151,6 +2151,46 @@ use super::*;
...
+
+        RECEIVED.store(false, Ordering::SeqCst);
+        let req_id = LOOKUP_REGISTRY.next_req_id();
+        LOOKUP_REGISTRY.lock().insert(req_id, Box::new(Callback));
+
+        let data = unsafe {
+            let data = libc::calloc(1, std::mem::size_of::<ffi::pmix_pdata_t>())
+                as *mut ffi::pmix_pdata_t;
+            assert!(!data.is_null());
+            (*data).value.type_ = PMIX_STRING as _;
```
SedahsReview: **Blocking — this still does not guard the double-free.** Here `PMIX_STRING` resolves to `crate::mock_ffi::PMIX_STRING = 2` (imported at tests.rs:1706), identical to the `PMIX_STRING_U16 = 2` you replaced. The real OpenPMIx `PMIX_STRING` is **3** (`pmix_common.h:1764`), and `free_value` (lib.rs:2905) matches the real constant. So this value's type (2) falls through `free_value` to `_ => {}` and is never freed by Rust; real `PMIx_Pdata_free` treats type 2 as scalar `PMIX_BYTE` with nothing to free. No double-free can occur in either branch — empirically the committed test passes (valgrind 0 errors) even with the bridge zeroing removed, while switching this line to `crate::ffi::PMIX_STRING` makes it SIGABRT on double-free. Use the real bindgen constant (value 3) so the test actually exercises the guard.

SedahsReview (changes requested):
## Re-review — Request Changes

**Verdict: the regression test still does not genuinely guard the double-free.** The core fix is correct, but the test change in commit ae1ef59 is effectively a no-op.

### What is correct (verified)
- `lookup_callback_bridge` zeroing (src/data_ops/mod.rs:776-779) exactly mirrors the blocking `lookup()` path (mod.rs:659-663): after `ptr::read(&pdata_ref.value)`, the source is `write_bytes`-zeroed and `type_` set to `PMIX_UNDEF` before `PMIx_Pdata_free`. This is the right ownership-transfer discipline.
- The `// SAFETY` comment (mod.rs:776-777) was added, resolving the prior non-blocking suggestion.
- The blocking `lookup()` path is unchanged.
- Real `PMIx_Pdata_free` (verified via disassembly of libpmix 6.1.0) does call `value_destruct` on each element and then `free()` — so the fix addresses a real double-free.

### Blocking issue — the test still does not catch the regression
The test sets `(*data).value.type_ = PMIX_STRING as _` (tests.rs:2178). Inside the test module, `PMIX_STRING` is imported from `crate::mock_ffi` (tests.rs:1706), where `mock_ffi::PMIX_STRING = 2` — identical to the old `PMIX_STRING_U16 = 2`. The real OpenPMIx `PMIX_STRING` is **3** (`pmix_common.h:1764`). `free_value` (lib.rs:2905) matches the real constant, so the value the test writes (type 2) falls through to `_ => {}` and is never freed by Rust — and real `PMIx_Pdata_free` treats type 2 as scalar `PMIX_BYTE` with no payload to free. So no double-free can occur in either branch.

**Empirical proof (run locally):**
- Committed test + fix removed → test passes, valgrind `0 errors`. The test does NOT abort.
- Test changed to `crate::ffi::PMIX_STRING` (=3) + fix removed → `free(): double free detected in tcache 2` (SIGABRT). The guard is real — but only with the *real* constant.

So the switch from `PMIX_STRING_U16` to `PMIX_STRING` changed nothing functionally in this scope (both resolve to 2). The prior blocking issue is not resolved.

### Required fix
Use the real FFI constant so `free_value` actually frees the payload: `(*data).value.type_ = crate::ffi::PMIX_STRING as _;` (or disambiguate the import so the test's `PMIX_STRING` is the bindgen constant, value 3). After that, reverting the bridge zeroing must make the test abort on double-free — that is the regression guard you want.

### Quality gates
- `cargo build` — clean.
- `cargo test --lib data_ops::tests -- --test-threads=1` — 277 passed, 0 failed.
- `cargo clippy --lib` — only pre-existing bindgen `missing_safety_doc` warnings; none from PR lines.
- `cargo fmt --check` — repo-wide pre-existing diffs; the fmt diff around mod.rs:777-782 is pre-existing context (the `Some(PmixOwnedValue ...)` line is a context line, not a PR addition), not a new PR-caused diff.

Reviewed against base `1207a255` at head `ae1ef59`.

SedahsReview:
## Code Review Summary — Changes Requested

**Verdict: REQUEST_CHANGES** (1 blocking issue) · reviewed at head `ae1ef59` vs base `1207a255`

### What looks good
- The **core fix is correct**: `lookup_callback_bridge` now zeroes the source pdata value (`write_bytes` + `type_ = PMIX_UNDEF`) after `ptr::read` ownership transfer, exactly mirroring the blocking `lookup()` path (mod.rs:659-663), before `PMIx_Pdata_free`. Real `PMIx_Pdata_free` (verified in libpmix 6.1.0 disassembly) does destruct each element + `free()`, so this addresses a genuine double-free.
- The requested `// SAFETY` comment was added above the zeroing lines.
- Blocking `lookup()` path is unmodified.

### 🔴 Blocking
- **src/data_ops/tests.rs:2178** — the regression test still does not catch the double-free. `PMIX_STRING` here resolves to `crate::mock_ffi::PMIX_STRING = 2` (imported tests.rs:1706), identical to the old `PMIX_STRING_U16 = 2`. Real OpenPMIx `PMIX_STRING = 3`, which is what `free_value` (lib.rs:2905) matches. With type 2 the payload is never freed by Rust nor by PMIx (scalar `PMIX_BYTE`), so no double-free can occur in either branch.
  - **Empirical proof:** committed test + fix removed → passes (valgrind 0 errors); test switched to `crate::ffi::PMIX_STRING` + fix removed → `double free detected` SIGABRT. The switch from `PMIX_STRING_U16` to `PMIX_STRING` was effectively a no-op.
  - **Fix:** use the real bindgen constant (value 3), e.g. `crate::ffi::PMIX_STRING as _`, so the test genuinely exercises the guard.

### Quality gates
- `cargo build` — clean
- `cargo test --lib data_ops::tests -- --test-threads=1` — 277 passed / 0 failed
- `cargo clippy --lib` — only pre-existing bindgen `missing_safety_doc` warnings; none from PR lines
- `cargo fmt --check` — repo-wide pre-existing diffs; the mod.rs:777-782 region is pre-existing context, not a new PR-caused diff

---

SedahsReview (approved):
## Verdict: ✅ APPROVE

Re-review of PR #455 after the prior blocking issue was addressed. The regression test now uses the **real** bindgen `PMIX_STRING` constant, making the double-free guard genuine.

### Fix verified
- **Correct constant**: `src/data_ops/tests.rs:2178` now writes `(*data).value.type_ = crate::ffi::PMIX_STRING as _;`. `crate::ffi` is the bindgen module (`include!(OUT_DIR/bindings.rs)`), and the generated constant is `pub const PMIX_STRING: u32 = 3;` — the real PMIx data-type value, **not** the `mock_ffi` value `2` (`PMIX_STRING_U16`). The prior blocking issue is resolved.
- **Genuine guard**: With `type_ = PMIX_STRING` (3), `free_value` (`src/lib.rs:2905`) hits the string arm and frees the `strdup`'d payload via `CString::from_raw` when the transferred `PmixOwnedValue` drops. I empirically verified the guard by temporarily reverting the bridge zeroing — the test aborts with `free(): double free detected in tcache 2` (SIGABRT). The regression test is real and will catch a regression of this fix.
- **Bridge source-zeroing** (`src/data_ops/mod.rs:776-779`) is present and correct: `ptr::read` transfers ownership of the value into `PmixOwnedValue`, then `write_bytes(&mut pdata_ref.value, 0, 1)` + `type_ = pmix_undef` clears the source **before** `PMIx_Pdata_free(data, ndata)` runs (line 796), so PMIx's destructor won't free the same payload twice. The SAFETY comment documents the ownership reasoning. This mirrors the blocking path.
- **Blocking `lookup()` path unchanged**; only `src/data_ops/mod.rs` and `src/data_ops/tests.rs` are touched. No leftover diagnostic probe files (`tests/zz_probe_doublefree.rs`, `/tmp/doublefree_probe.rs` absent).

### Quality gates (run on head `9cbeb5a`)
- `cargo build` — ✅ clean
- `cargo test --lib data_ops::tests::test_lookup_callback_bridge* -- --test-threads=1` — ✅ 6/6 pass (incl. the new regression test)
- `cargo clippy --lib -- -D warnings` — ✅ no findings on PR lines (only pre-existing `# Safety`-doc errors in generated `bindgen` output)
- `cargo fmt --check` — ✅ no new diffs from PR lines (pre-existing repo diffs only)

No remaining blocking issues. Approving.

SedahsReview:
## Code Review Summary — ✅ APPROVED (round 3, review 4912722505)

**Verdict: APPROVE.** The prior blocking issue is resolved.

### What was fixed
The double-free regression test now writes `(*data).value.type_ = crate::ffi::PMIX_STRING as _;` — the **real bindgen** constant (value **3**), not the `mock_ffi` `PMIX_STRING`/`PMIX_STRING_U16` (value 2). With value 3, `free_value` frees the `strdup`'d payload, so the guard is genuine. I confirmed empirically: reverting the bridge zeroing makes the test abort with `free(): double free detected` (SIGABRT).

### Verified on head `9cbeb5a`
- `cargo build` ✅
- `cargo test --lib data_ops::tests::test_lookup_callback_bridge* -- --test-threads=1` — 6/6 pass ✅
- `cargo clippy --lib -- -D warnings` — no findings on PR lines (pre-existing bindgen `# Safety` errors only) ✅
- `cargo fmt --check` — no new diffs from PR lines (pre-existing repo diffs only) ✅
- Blocking `lookup()` path unchanged; no leftover probe files.

### Remaining issues
None (blocking). Non-blocking note: `pdata_ref.value.type_ = pmix_undef` at mod.rs:779 is redundant since the preceding `write_bytes(...,0,1)` already zeroes it, but it is explicit and harmless — no change required.

---

PR #461, opened 2026-08-12
fix(groups,server): call release_fn in group/collect_inventory nb completions instead of PMIx_Info_free

### LINKED ISSUE #442: high: group _nb completions (and collect_inventory) ignore release_fn — leak a PMIx tracker per call
## Severity: high (leak per group op + latent double-free)

`src/groups.rs:230-266, 430-466, 632-668` (group_construct/invite/join bridges) and `src/server/mod.rs:2394-2395`.

For `PMIx_Group_{construct,invite,join}_nb`, OpenPMIx invokes the user's `pmix_info_cbfunc_t` as `cbfunc(ret, cb->info, cb->ninfo, cb->cbdata, relfn, cb)` (pmix_client_group.c:1372, 1186) — the info array remains owned by the `pmix_group_tracker_t`, and the recipient **must** call `release_fn(release_cbdata)` when finished; it is the only path that releases the tracker. The Rust bridges take `info` into `GroupResults` (freed via `PMIx_Info_free` on drop) and never call `release_fn`: each group op permanently leaks its tracker, and the info array is freed underneath the still-referencing tracker (latent double-free). Same defect in `collect_inventory_callback_bridge` (`let _ = release_fn;` while also freeing with `PMIx_Info_free`). On error status with non-null info, the group bridges also drop the info pointer without freeing — additional leak.

**Fix:** store `(release_fn, release_cbdata)` in the result object and call `release_fn` on drop instead of `PMIx_Info_free`; free-or-release info on error paths too.

SedahsDev:
Closes #442

## Summary
Release PMIx-owned tracker-backed info arrays through the completion release callback instead of PMIx_Info_free.

## Affected functions
Updated group_construct_callback_bridge, group_invite_callback_bridge, group_join_callback_bridge, and collect_inventory_callback_bridge, including missing-callback/error cleanup. Blocking group result ownership remains unchanged.

## Rationale
OpenPMIx owns these arrays through group/inventory trackers; release_fn(release_cbdata) is the sole operation that releases both tracker and array.

## Test plan
- Local mock-based unit regression verifies group completion invokes release_fn with release_cbdata when GroupResults is dropped.
- cargo check --lib --offline passed with PMIX_PREFIX=$HOME/.local/openpmix-6.1.0.
- Full daemon-based tests require CI DVM/prterun and are not available locally.

### DIFF
--- a/src/groups.rs
+++ b/src/groups.rs
@@ -74,12 +74,17 @@ fn flat_infos(infos: &[Info]) -> Vec<ffi::pmix_info_t> {
 
 /// Owned result set from a group operation.
 ///
-/// Wraps the `pmix_info_t*` array returned by `PMIx_Group_*` and frees it once
-/// via `PMIx_Info_free` on drop. The array is one allocation, not one per entry.
+/// Result set from a group operation.
+///
+/// Blocking calls own their returned array and release it with
+/// `PMIx_Info_free`. Non-blocking calls view a PMIx-owned tracker array and
+/// retain the release callback that frees the tracker and array.
 #[derive(Debug)]
 pub struct GroupResults {
     handle: *mut ffi::pmix_info_t,
     len: usize,
+    release_fn: Option<unsafe extern "C" fn(*mut c_void)>,
+    release_cbdata: *mut c_void,
     /// Makes this type `!Send` + `!Sync` (owns PMIx/C memory — not free-threaded).
     _not_thread_safe: std::marker::PhantomData<*mut u8>,
 }
@@ -95,7 +100,13 @@ impl GroupResults {
 
 impl Drop for GroupResults {
     fn drop(&mut self) {
-        if !self.handle.is_null() && self.len > 0 {
+        if let Some(release_fn) = self.release_fn.take() {
+            // SAFETY: PMIx supplied this callback and opaque data for this
+            // completion; it releases the tracker-owned info array.
+            unsafe { release_fn(self.release_cbdata) };
+            self.handle = ptr::null_mut();
+            self.len = 0;
+        } else if !self.handle.is_null() && self.len > 0 {
             #[cfg(any(test, feature = "mock_ffi"))]
             if crate::mock_ffi::is_mock_enabled() {
                 // SAFETY: handle is the single allocation returned by PMIx_Group_*.
@@ -194,6 +205,8 @@ pub fn group_construct(
         } else {
             nresults
         },
+        release_fn: None,
+        release_cbdata: ptr::null_mut(),
         _not_thread_safe: std::marker::PhantomData,
     })
 }
@@ -232,8 +245,8 @@ pub unsafe extern "C" fn group_construct_callback_bridge(
     info: *mut ffi::pmix_info_t,
     ninfo: usize,
     cbdata: *mut c_void,
-    _release_fn: Option<unsafe extern "C" fn(*mut c_void)>,
-    _release_cbdata: *mut c_void,
+    release_fn: Option<unsafe extern "C" fn(*mut c_void)>,
+    release_cbdata: *mut c_void,
 ) {
     if cbdata.is_null() {
         return;
@@ -244,6 +257,9 @@ pub unsafe extern "C" fn group_construct_callback_bridge(
         registry.remove(&req_id)
     };
     let Some(cb_wrapper) = cb else {
+        if let Some(release_fn) = release_fn {
+            unsafe { release_fn(release_cbdata) };
+        }
         return;
     };
     let pmix_status = PmixStatus::from_raw(status);
@@ -258,6 +274,8 @@ pub unsafe extern "C" fn group_construct_callback_bridge(
         } else {
             0
         },
+        release_fn,
+        release_cbdata,
         _not_thread_safe: std::marker::PhantomData,
     };
     let _ = invoke_user_callback("groups", move || {
@@ -398,6 +416,8 @@ pub fn group_invite(
         } else {
             nresult
         },
+        release_fn: None,
+        release_cbdata: ptr::null_mut(),
         _not_thread_safe: std::marker::PhantomData,
     })
 }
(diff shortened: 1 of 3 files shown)


src/groups.rs
```diff
@@ -244,6 +257,9 @@ pub unsafe extern "C" fn group_construct_callback_bridge(
         registry.remove(&req_id)
     };
     let Some(cb_wrapper) = cb else {
+        if let Some(release_fn) = release_fn {
+            unsafe { release_fn(release_cbdata) };
```
SedahsReview, line 261: Minor (non-blocking): this `unsafe { release_fn(release_cbdata) };` in the missing-callback path (and the identical ones at groups.rs:468, 677 and server/mod.rs:2376) lacks a `// SAFETY:` comment, unlike the Drop impls which document the call. Suggest adding one for consistency, e.g. `// SAFETY: PMIx supplied this callback/opaque data for this completion; it releases the tracker-owned info array.`

src/server/mod.rs
```diff
@@ -2371,11 +2371,9 @@ pub(crate) extern "C" fn collect_inventory_callback_bridge(
...
         Some(cb) => cb,
         None => {
-            // Callback already consumed — free the info array to avoid leak.
-            if !info.is_null() && ninfo > 0 {
-                unsafe {
-                    ffi::PMIx_Info_free(info, ninfo);
-                }
+            // Callback already consumed — release the PMIx tracker.
+            if let Some(release_fn) = release_fn {
+                unsafe { release_fn(release_cbdata) };
```
SedahsReview, line 2376: Minor (non-blocking): same missing `// SAFETY:` comment as noted on groups.rs:261 — this `release_fn` call in the already-consumed-callback path is safe (PMIx-supplied release callback/opaque data) but undocumented. Add a `// SAFETY:` line for consistency with the `CollectInventoryResults` Drop impl.

SedahsReview (approved):
## Code Review — PMIx tracker release fix (#442)

**Verdict: APPROVE** — the ownership fix is correct, safe, and well-tested. Two minor, non-blocking suggestions below.

### What this PR gets right
The root cause of #442 was that the `_nb` completion bridges treated the PMIx-owned tracker info array as Rust-owned (freeing it via `PMIx_Info_free`) and never released the underlying tracker. This PR correctly switches to `release_fn(release_cbdata)` as the sole release path for non-blocking results.

- **`GroupResults`** now carries `release_fn` + `release_cbdata`. `Drop` calls `release_fn` (via `.take()`, so exactly once) with the correct `release_cbdata` and nulls `handle`/`len` — it does **not** `PMIx_Info_free` the tracker-owned array, so no double-free. The blocking-path `PMIx_Info_free` fallback is preserved for `release_fn: None`.
- **All three group bridges** (construct/invite/join) store `(release_fn, release_cbdata)` into the result object on every path — including error status (where `handle` is nulled but `release_fn` is retained, so the tracker is still released on drop) — and the missing-callback path now calls `release_fn` directly instead of leaking.
- **`collect_inventory_callback_bridge`**: `let _ = release_fn;` is gone. `release_fn(release_cbdata)` is invoked both on the missing-callback path and from `CollectInventoryResults::Drop`; no `PMIx_Info_free` on the tracker-owned array. `server_collect_inventory` is the only constructor and is non-blocking, so there's no blocking-path leak regression.
- **Public API is unchanged** — the user drops the result object and the tracker is released for them; nothing else needs freeing.

### Verification
- `cargo build` — passes (PMIX_PREFIX=$HOME/.local/openpmix-6.1.0).
- `cargo test --lib groups::tests -- --test-threads=1` — 96 passed.
- `cargo test --lib server::tests -- --test-threads=1` — 234 passed.
- New regression test `group_construct_bridge_releases_tracker_on_result_drop` passes and deterministically verifies `release_fn` is invoked exactly once with the passed `release_cbdata` (the user callback runs synchronously via `invoke_user_callback`, so the drop happens before the assert).
- `cargo clippy --lib` — only the pre-existing bindgen `missing_safety_doc` warnings; nothing from the PR's added lines.
- `cargo fmt --check` — repo has pre-existing diffs; one new diff from the PR's added test (see suggestion #2).

- Add `// SAFETY:` comments to the four `unsafe { release_fn(release_cbdata) };` calls in the missing-callback/error paths (groups.rs:261, 468, 677; server/mod.rs:2376) for consistency with the `Drop` impls, which already document why the call is safe.
- rustfmt the new test's `GROUP_CONSTRUCT_REGISTRY.lock().insert(...)` in groups.rs:1035.

No critical or warning-level issues found.

SedahsReview:
## Review Summary — Hermes Agent

**Verdict: APPROVE** ✅ (review id 4913433660)

### 🔴 Critical
- None.

### ⚠️ Warnings
- None.

### 💡 Suggestions (non-blocking)
- `src/groups.rs:261, 468, 677` and `src/server/mod.rs:2376` — the error/missing-callback `unsafe { release_fn(release_cbdata) };` blocks lack `// SAFETY:` comments, inconsistent with the `Drop` impls. Add them for documentation consistency.
- `src/groups.rs:1035` — new test's `GROUP_CONSTRUCT_REGISTRY.lock().insert(...)` isn't rustfmt-formatted.

### ✅ Looks Good
- **Correct ownership fix for #442.** `GroupResults` and `CollectInventoryResults` now release the PMIx tracker via `release_fn(release_cbdata)` instead of `PMIx_Info_free` on the tracker-owned array — no tracker leak, no double-free. `release_fn` is invoked exactly once (`.take()` in `Drop`), with the correct `release_cbdata`; `handle` is nulled and never `PMIx_Info_free`d on the non-blocking path.
- **All three group bridges** (construct/invite/join) store `(release_fn, release_cbdata)` on every path including error status, and the missing-callback path calls `release_fn` directly — no leaked tracker.
- **`collect_inventory_callback_bridge`**: `let _ = release_fn;` removed; `release_fn(release_cbdata)` called on both the consumed-callback path and via `CollectInventoryResults::Drop`. `server_collect_inventory` is the only constructor (non-blocking), so no blocking-path leak regression.
- **Public API unchanged** — user just drops the result object; tracker is released for them.

### Build / Test / Lint / Format
- `cargo build` — ✅ passes (PMIX_PREFIX=$HOME/.local/openpmix-6.1.0)
- `cargo test --lib groups::tests -- --test-threads=1` — ✅ 96 passed
- `cargo test --lib server::tests -- --test-threads=1` — ✅ 234 passed
- New test `group_construct_bridge_releases_tracker_on_result_drop` — ✅ passes
- `cargo clippy --lib` — ✅ only pre-existing bindgen `missing_safety_doc` warnings; nothing from PR's added lines
- `cargo fmt --check` — ⚠️ repo has pre-existing diffs; one new diff from the PR's added test (suggestion above)

---

PR #488, opened 2026-08-13
fix(process_mgmt): flatten Proc array for PMIx_Abort (slice-stride fix)

### LINKED ISSUE #474: abort() passes &Proc[0].handle with procs.len() as if Proc were a pmix_proc_t[] (slice-stride garbage for n>1)
**Severity:** CRITICAL (wrong data / memory misread for multi-proc)  **Source:** Claude Opus, GLM

`process_mgmt.rs::abort` builds the proc pointer as:

```rust
&procs[0].handle as *const ffi::pmix_proc_t, procs.len()
```

But `Proc` is `{ handle: pmix_proc_t, len: usize }` (default repr) — `sizeof(Proc) != sizeof(pmix_proc_t)`. PMIx strides by `sizeof(pmix_proc_t)` across elements that are `sizeof(Proc)` apart, so every proc after index 0 is read at the wrong offset (garbage nspace/rank) for `nprocs > 1`. The crate already has `flat_procs()` in the same file (used by `connect`/`disconnect`/`spawn`), so this is an easy fix.

**Fix:** `let flat = flat_procs(procs); PMIx_Abort(status, msg, flat.as_ptr(), flat.len())`.

- src/process_mgmt.rs:129-153

SedahsDev:
Closes #474

## Summary
- abort() now uses flat_procs() to build a contiguous Vec<pmix_proc_t>, kept alive for the duration of the blocking PMIx_Abort call, instead of passing &Proc[0].handle with procs.len() (which mis-strides for n>1 since sizeof(Proc) != sizeof(pmix_proc_t)).

## Test plan
- cargo build, cargo test --lib, cargo clippy --lib, cargo check --features mock_ffi --lib.

### DIFF
--- a/src/process_mgmt.rs
+++ b/src/process_mgmt.rs
@@ -125,12 +125,13 @@ pub fn abort(
         None => (ptr::null(), None),
     };
 
-    // Convert the optional proc array to a raw pointer + length.
+    // Convert the optional proc array to a contiguous raw pointer + length.
+    let flat;
     let (procs_ptr, nprocs) = match procs {
-        Some(procs) if !procs.is_empty() => (
-            &procs[0].handle as *const ffi::pmix_proc_t as *mut ffi::pmix_proc_t,
-            procs.len(),
-        ),
+        Some(procs) if !procs.is_empty() => {
+            flat = flat_procs(procs);
+            (flat.as_ptr() as *mut ffi::pmix_proc_t, flat.len())
+        }
         _ => (ptr::null_mut(), 0),
     };
 
@@ -140,9 +141,9 @@ pub fn abort(
     // - `msg_ptr` is either null or a valid NUL-terminated C string
     //   whose lifetime (`_msg_cstring`) is kept alive until after this
     //   call returns.
-    // - `procs_ptr` is either null or points to a valid slice of
-    //   `pmix_proc_t` handles that remain valid for the duration of
-    //   this call. PMIx does not retain these pointers after return.
+    // - `procs_ptr` is either null or points to a temporary flattened
+    //   `Vec<pmix_proc_t>` that remains alive for the duration of this call.
+    //   PMIx does not retain these pointers after return.
     // - Note: if the caller's own process is included in `procs`, this
     //   function may not return. That is documented PMIx behavior.
     let raw_status = unsafe {


SedahsReview (approved):

Reviewed PR #488 (fix/abort-flat-procs → main), closes #474. Single-file change to `src/process_mgmt.rs` (+9/-8).

### Correctness (#474)
- `flat_procs(procs)` builds a contiguous `Vec<ffi::pmix_proc_t>` by copying each `Proc.handle` by value, so the C array stride matches `sizeof(pmix_proc_t)`. This correctly fixes the prior mis-striding for n>1 (where `sizeof(Proc) != sizeof(pmix_proc_t)`).
- `let flat;` is declared at function scope and only assigned in the non-empty arm; because it is a top-level binding, the Vec stays alive until function exit — well past the blocking `PMIx_Abort` call at line 149. Deferred initialization is sound: `flat` is only read on the path where it was assigned.
- `flat.as_ptr() as *mut ffi::pmix_proc_t` is sound: the Vec is owned and alive for the whole call, and PMIx does not retain the pointer after return. Casting `*const`→`*mut` is appropriate since the C API takes a non-const array but does not mutate it.
- `flat.len()` is passed as `nprocs` — correct.

### Scope / hygiene
- Change is append-only within the function; no unrelated edits, secrets, debug code, or merge markers.
- SAFETY comment updated to accurately describe the flattened-Vec lifetime.

### Test coverage
- `test_abort_with_multiple_procs` exercises the n>1 path that the old code got wrong (passes). All 9 abort-related tests pass under `--test-threads=1`.
- Note: these tests assert only an error return (no DVM/daemon locally), so they validate the parameter path but cannot assert the C-level stride directly — that's CI-only, and the fix is correct from inspection.

### Gates (verified independently)
- `cargo build` — clean.
- `cargo check --features mock_ffi --lib` — passes, 45 baseline warnings.
- `cargo test --lib -- --test-threads=1 abort` — 9 passed, 0 failed.

No blocking issues found. Approving.

---

SedahsReview:
## Code Review Summary — PR #488

**Verdict: APPROVED** (review id 4922895047)

### Correctness (fixes #474)
- `flat_procs` produces a contiguous `Vec<pmix_proc_t>` by value-copying each `Proc.handle` — correct array stride for n>1 (the old `&procs[0].handle` + `procs.len()` mis-strided because `sizeof(Proc) != sizeof(pmix_proc_t)`).
- `let flat;` at function scope keeps the Vec alive through the blocking `PMIx_Abort` call; dropped after return.
- Pointer cast and `flat.len()` (nprocs) both sound/correct.

### Scope & hygiene
- Append-only (+9/-8, 1 file), SAFETY comment updated accurately, no unrelated changes/secrets/merge markers.

### Gates (independently verified)
- `cargo build` clean
- `cargo check --features mock_ffi --lib` passes (45 baseline warnings)
- `cargo test --lib -- --test-threads=1 abort` — 9 passed / 0 failed, incl. `test_abort_with_multiple_procs` (n>1)

### Note (non-blocking)
- `test_abort_with_multiple_procs` only asserts an error return (no DVM locally); the C-level stride can't be asserted without a daemon — CI-only. Fix is correct from inspection.

---

PR #489, opened 2026-08-13
fix(utility): iof_pull_blocking returns real refid via async reg callback

### LINKED ISSUE #476: iof_pull_blocking uses the blocking return status as the registration handle — registry never matches
**Severity:** HIGH (broken API)  **Source:** GPT-5.6, Claude Opus, Grok

`utility/mod.rs::iof_pull_blocking` does `let handle = raw_result as usize;` after a blocking `PMIx_IOF_pull` (regcbfunc=NULL). But OpenPMIx's blocking mode returns `PMIX_OPERATION_SUCCEEDED` (a status), **never** the registration id — the refid is only delivered through the registration callback (`myregcbfunc` → `req->local_id`). The wrapper keys `IOF_REGISTRY` by the bogus status value, so `io_callback_bridge` (keyed by the real refid) never matches and `iof_deregister(handle)` targets the wrong id.

**Fix:** Implement blocking mode by driving the async path internally: register with a real reg-callback, block on a channel/condvar until the bridge delivers the actual refid, and key the registry with that refid.

- src/utility/mod.rs:1148-1207

SedahsDev:
Closes #476

## Summary
- iof_pull_blocking now drives the async path internally: it passes the real reg_callback_bridge and blocks on a channel until the bridge delivers the actual refid, then keys IOF_REGISTRY by that refid and returns it. Previously it keyed the registry by the blocking-mode return STATUS, so io_callback_bridge never matched and iof_deregister targeted the wrong id.

## Test plan
- cargo build, cargo test --lib, cargo clippy --lib, cargo check --features mock_ffi --lib.

### DIFF
--- a/src/utility/mod.rs
+++ b/src/utility/mod.rs
@@ -1145,7 +1145,8 @@ where
 /// * `Err(PmixStatus)` — registration failed.
 ///
 /// # C API
-/// Same as `PMIx_IOF_pull` with `regcbfunc` set to NULL (blocking mode).
+/// Same as `PMIx_IOF_pull`, using the asynchronous registration callback
+/// internally to obtain the real registration handle.
 pub fn iof_pull_blocking<F>(
     procs: &[ffi::pmix_proc_t],
     directives: &[ffi::pmix_info_t],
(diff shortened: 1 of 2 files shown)


src/utility/mod.rs
```diff
@@ -1155,54 +1156,62 @@ pub fn iof_pull_blocking<F>(
...
-        }
-
-        // Store the context in the registry so the IO callback can find it.
-        {
-            let mut registry = IOF_REGISTRY.lock().unwrap();
-            registry.insert(handle, ctx);
+        // reg_callback_bridge reclaimed ctx_ptr and inserted the real refid.
+        let (status, refid) = rx
+            .recv()
+            .expect("PMIx_IOF_pull succeeded without invoking its registration callback");
```
SedahsReview, line 1208: Non-blocking suggestion: `rx.recv()`'s `.expect()` would panic (unwinding across the FFI boundary is contained by `invoke_user_callback`, but the panic surfaces to the caller) if the channel disconnected without a send. Per the OpenPMIx contract this can't happen on a non-error `PMIx_IOF_pull` return (myreg/myregcbfunc always invoke regcbfunc after a successful registration), so the expect is defensible. A more defensive option would be `recv().map_err(|_| ...)` or `recv_timeout` to fail gracefully instead of panicking if the daemon misbehaves. Not blocking.

SedahsReview (approved):

Reviewed `fix(utility): iof_pull_blocking returns real refid via async reg callback` (PR #489, closes #476).

**Summary:** The fix is correct and resolves the underlying bug. Previously `iof_pull_blocking` used blocking mode (`regcbfunc = None`), in which OpenPMIx returns `PMIX_OPERATION_SUCCEEDED` from the call and the real registration handle never reaches the caller — the return value was taken as the handle. The new implementation routes through the async registration callback to obtain the real refid, which is the correct handle to key the registry by.

**Ownership/`ctx_ptr` (verified against `pmix_iof.c`):**
- Error path: on an immediate FFI error, OpenPMIx returns before threadshift, so `regcbfunc` is never invoked (verified: INIT/NOT_AVAILABLE/NOT_SUPPORTED/NOMEM/UNREACH all return before `PMIX_THREADSHIFT`). Reclaiming `ctx_ptr` with `Box::from_raw` is therefore correct.
- Success path: `reg_callback_bridge` reclaims the temp Arc box at its top (`Box::from_raw(cbdata)`) and inserts `Arc::clone(&ctx)` into `IOF_REGISTRY` keyed by the real refid. `iof_pull_blocking` does **not** also reclaim `ctx_ptr`, so there is no double-free. Confirmed the blocking fn no longer calls `Box::from_raw(ctx_ptr)` on the success branch.

**Blocking semantics:** `rx.recv()` blocks until `reg_callback_bridge` (running on the progress thread via `invoke_user_callback`) sends `(status, refid)`. This is a genuine blocking call as the contract requires; the channel send is guaranteed on a non-error FFI return per the C code (myreg/myregcbfunc always call regcbfunc after a successful registration).

**Registry correctness:** Keyed by the real refid so `io_callback_bridge` lookups and `iof_deregister(refid)` target the right entry; on async-error status the registry entry is removed before returning `Err`, so no leak.

**Mock integration:** `mock_iof_pull` (unchanged) invokes the reg callback synchronously with the configured handle when `regcbfunc` is `Some`, so the new test exercises the full path: mock handle 42 → `Ok(42)` and `IOF_REGISTRY` contains key 42. Meaningful assertion.

**Append-only:** The async `iof_pull` and `reg_callback_bridge` are unchanged (0 hunks in `pub fn iof_pull<` / `extern "C" fn reg_callback_bridge`). Only `iof_pull_blocking` and the test changed. No secrets, debug statements, or merge markers.

**Gates (run on PR branch):** `cargo build` clean; `cargo check --features mock_ffi --lib` passes (45 baseline warnings, unchanged); `test_iof_pull_blocking_returns_real_refid` passes; full `iof` test subset 37 passed / 0 failed; utility module 204 passed / 0 failed.

One non-blocking suggestion (see inline comment) about `recv()`'s `.expect()` vs. graceful error handling. Otherwise the change is correct and ready to merge.

SedahsReview:
## Code Review Summary — APPROVED ✅

**PR #489** `fix(utility): iof_pull_blocking returns real refid via async reg callback` (closes #476)

**Verdict: APPROVE** — clean, correct concurrency/ownership handling. Full formal review posted inline.

### What was verified
- **Ownership of `ctx_ptr`:** Error path reclaims correctly (`Box::from_raw`) since OpenPMIx returns before threadshift on immediate errors (regcbfunc never fires — verified in `pmix_iof.c`). Success path leaves reclamation to `reg_callback_bridge`; **no double-free**.
- **Blocking semantics:** `rx.recv()` blocks until the reg callback delivers the real refid on the progress thread — guaranteed on non-error FFI return per C contract.
- **Registry:** keyed by real refid; async-error status removes the entry before returning `Err`. No leak.
- **Mock integration:** `mock_iof_pull` (unchanged) drives the reg callback synchronously with handle 42; test meaningfully asserts `Ok(42)` + registry key 42.
- **Append-only:** async `iof_pull` / `reg_callback_bridge` unchanged.

### Gates
- `cargo build` clean; `cargo check --features mock_ffi --lib` passes (45 baseline warnings); `test_iof_pull_blocking_returns_real_refid` passes; iof subset 37/0; utility module 204/0.

### Minor (non-blocking)
- `src/utility/mod.rs:1208` — `rx.recv().expect(...)` panics if the channel disconnects; `map_err`/`recv_timeout` would be more defensive. Optional.

PR #504, opened 2026-08-24
fix(data_ops): retain get_nb key+info across threadshift

### LINKED ISSUE #491: high: get_nb passes a local CString key and borrowed Info into PMIx_Get_nb which aliases both across its threadshift (UAF)
## Description
`data_ops::get_nb` (src/data_ops/mod.rs:316) builds `key_c` as a **local** `CString` and takes `info: Option<&Info>` as a borrow, passes `key_c.as_ptr()` / `info.handle` to `PMIx_Get_nb`, then returns `Ok(())`. The caller can drop the key string and the `Info` immediately after.

OpenPMIx aliases both inputs across its threadshift:

- src/client/pmix_client_get.c:351 — `cb->key = (char*)key;`
- :352-358 region — the cb object stores `info`/`ninfo` verbatim
- :359 — `PMIX_THREADSHIFT(cb, get_data);`

`get_data` runs on the progress thread **after** `PMIx_Get_nb` returns and dereferences both the key pointer and the info array (`cb->key` is packed into the request at :551). If the caller drops anything before delivery, this is use-after-free.

## Fix
Retain an owned copy of the key (`CString` moved into a per-req registry entry) and deep-copy the info array (like allocation_request_nb already does with `into_boxed_slice()` + retained-info registry); free in `get_value_callback_bridge` on all completion paths including the immediate-error branch.

Severity: high. Every non-blocking get with a temporary key (the common case) is affected.

SedahsDev:
Closes #491

## Summary
- get_nb now retains an owned copy of the key CString and a deep-copied info array in the per-request registry, freed in get_value_callback_bridge on all completion paths — fixes UAF where PMIx_Get_nb aliases both across its threadshift.

## Test plan
- cargo build, cargo test --lib (full suite), cargo clippy --lib, cargo check --features mock_ffi --lib.

### DIFF
--- a/src/data_ops/mod.rs
+++ b/src/data_ops/mod.rs
@@ -206,8 +206,26 @@ pub trait GetValueCallback: Send {
 /// function uses this pointer to recover the Rust closure from the registry.
 type GetRegistry = Registry<Box<dyn GetValueCallback>>;
 static GET_REGISTRY: LazyLock<GetRegistry> = LazyLock::new(GetRegistry::new);
+type GetRetainedInputs = (CString, usize, usize);
+static GET_RETAINED_INPUTS: LazyLock<Registry<GetRetainedInputs>> =
+    LazyLock::new(Registry::new);
 static QUALIFIED_GETS: LazyLock<Mutex<std::collections::HashSet<usize>>> =
     LazyLock::new(|| Mutex::new(std::collections::HashSet::new()));
+
+fn release_retained_get_inputs(req_id: usize) {
+    if let Some((_key, address, len)) = GET_RETAINED_INPUTS.remove(req_id) {
+        if address != 0 {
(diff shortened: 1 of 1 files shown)


SedahsReview:
Review: **APPROVE** on b64a0a94. Verdict: APPROVE. get_nb now retains an owned copy of the key CString + deep-copied info in a per-req registry, freed on all completion/error paths — fixes UAF where PMIx_Get_nb aliases both across threadshift. No double-free (remove is single-shot). Full lib suite 1861 passed.

SedahsReview (approved):
Verdict: APPROVE. get_nb now retains an owned copy of the key CString + deep-copied info in a per-req registry, freed on all completion/error paths — fixes UAF where PMIx_Get_nb aliases both across threadshift. No double-free (remove is single-shot). Full lib suite 1861 passed.

